PR #636, opened 2024-11-15
Refactor dependency resolution out of deployment handler

sameerzuberi:
*Issue #, if available:*

*Description of changes:*
- Remove all dependency resolution logic from deployment_handler. This logic is now in a new dependency_resolver which will be called from within deployment_handler.
- Refactor all deployment configuration logic into deployment_configuration. This makes it easier to get any necessary config values and streamlines deployment handler code.


rawalexe:
PR #762 covers it

PR #658, opened 2024-11-26
Add dependency on time sync to networked services

cookpate:
*Issue #, if available:*

*Description of changes:*


cookpate:
Ah, chrony and systemd-time-wait-sync use differing protocols (NTP vs SNTP), so they are not aliased to each-other. Adding chrony to the `After=` list could be done.

thomas-roos:
If a service is not existing it will ignored or?

cookpate:
`After=` Is a weak dependency. It is ignored if the dependent service does not exist or is not starting.

PR #688, opened 2024-12-05
Support config merge and reset for cloud deployments

patrzhan:
*Issue #, if available:*

*Description of changes:*
- Add config merge reset functionality to cloud deployments. Deployments that specify a config merge or reset update will apply these changes to the component configurations.
- Delete the thing group to root components key from the database before writing. This ensures that in scenarios where components were removed, we don't erroneously pick those components back up if a different thing group deployment is made.

### DIFF
--- a/ggdeploymentd/src/component_config.c
+++ b/ggdeploymentd/src/component_config.c
@@ -1,0 +1,198 @@
+// aws-greengrass-lite - AWS IoT Greengrass runtime for constrained devices
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "component_config.h"
+#include "deployment_model.h"
+#include <assert.h>
+#include <ggl/buffer.h>
+#include <ggl/core_bus/gg_config.h>
+#include <ggl/error.h>
+#include <ggl/list.h>
+#include <ggl/log.h>
+#include <ggl/map.h>
+#include <ggl/object.h>
+#include <stdbool.h>
+#include <stddef.h>
+
+static GglError apply_reset_config(
+    GglBuffer component_name, GglMap component_config_map
+) {
+    GglObject *reset_configuration = NULL;
+    GglError ret = ggl_map_validate(
+        component_config_map,
+        GGL_MAP_SCHEMA(
+            { GGL_STR("reset"), false, GGL_TYPE_LIST, &reset_configuration }
+        )
+    );
+    if (ret != GGL_ERR_OK) {
+        return ret;
+    }
+
+    // If there is no reset configuration, then there is no
+    // configuration update to make
+    if (reset_configuration == NULL) {
+        return GGL_ERR_OK;
+    }
+
+    if (reset_configuration->type != GGL_TYPE_LIST) {
+        GGL_LOGE("Reset update did not parse into a list "
+                 "during configuration updates.");
+        return GGL_ERR_INVALID;
+    }
+    GGL_LIST_FOREACH(reset_element, reset_configuration->list) {
+        if (reset_element->type != GGL_TYPE_BUF) {
+            GGL_LOGE("Configuration key for reset config "
+                     "update not provided as a buffer.");
+            return GGL_ERR_INVALID;
+        }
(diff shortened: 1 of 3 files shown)


ggdeploymentd/src/deployment_handler.c
```diff
@@ -1977,9 +1990,188 @@ static GglError wait_for_deployment_status(GglMap resolved_components) {
...
+            { GGL_STR("reset"), false, GGL_TYPE_LIST, &reset_configuration }
+        )
+    );
+    if (ret != GGL_ERR_OK) {
+        return ret;
+    }
+
+    // If there is no reset configuration, then there is no
+    // configuration update to make
+    if (reset_configuration != NULL) {
```
sameerzuberi, line 2009: nit: Let's check for `reset_configuration == NULL` and then return if it does. Saves us from adding indentations to the code

ggdeploymentd/src/deployment_handler.c
```diff
@@ -1977,9 +1990,188 @@ static GglError wait_for_deployment_status(GglMap resolved_components) {
...
+            { GGL_STR("merge"), false, GGL_TYPE_MAP, &merge_configuration }
+        )
+    );
+    if (ret != GGL_ERR_OK) {
+        return ret;
+    }
+
+    // If there is no merge configuration, then there is no
+    // configuration update to make
+    if (merge_configuration != NULL) {
```
sameerzuberi, line 2066: Same as above, let's check for `merge_configuration == NULL` and then return

ggdeploymentd/src/deployment_handler.c
```diff
@@ -1977,9 +1990,188 @@ static GglError wait_for_deployment_status(GglMap resolved_components) {
...
+            { component_name, false, GGL_TYPE_MAP, &doc_component_info }
+        )
+    );
+    if (ret != GGL_ERR_OK) {
+        return ret;
+    }
+
+    // No config items to write if the component is not a root component in
+    // the deployment
+    if (doc_component_info != NULL) {
```
sameerzuberi, line 2124: Same as above

ggdeploymentd/src/deployment_handler.c
```diff
@@ -1977,9 +1990,188 @@ static GglError wait_for_deployment_status(GglMap resolved_components) {
...
+                             false,
+                             GGL_TYPE_MAP,
+                             &component_configuration })
+        );
+        if (ret != GGL_ERR_OK) {
+            return ret;
+        }
+
+        // No config items to write if there is no configurationUpdate item
+        if (component_configuration != NULL) {
```
sameerzuberi, line 2144: Here as well

ggdeploymentd/src/deployment_handler.c
```diff
@@ -1977,9 +1990,188 @@ static GglError wait_for_deployment_status(GglMap resolved_components) {
...
+            "Made a configuration merge update for component %.*s",
+            (int) component_name.len,
+            component_name.data
+        );
+    }
+
+    return GGL_ERR_OK;
+}
+
+static GglError apply_configurations(
```
sameerzuberi, line 2103: Can we move this method and the merge and reset methods into a separate file, maybe something like `config_update_handler`? Adding to deployment handler makes it more difficult to manage, this will save us from some refactoring in the future.

sameerzuberi (approved):

ggdeploymentd/src/component_config.c
```diff
@@ -0,0 +1,198 @@
...
+    if (reset_configuration == NULL) {
+        return GGL_ERR_OK;
+    }
+
+    if (reset_configuration->type != GGL_TYPE_LIST) {
+        GGL_LOGE("Reset update did not parse into a list "
+                 "during configuration updates.");
+        return GGL_ERR_INVALID;
+    }
+    GGL_LIST_FOREACH(reset_element, reset_configuration->list) {
```
aws-kevinrickard, line 43: what do these `reset_element`s look like?

If I want to reset a key c with path a/b/c , wouldn't we need to parse that path into a list like [ a , b , c ] to pass to the config delete like [ services , component , configuration , a , b , c ] ?

It seems like since we're only using it as a buffer you can only reset non-nested keys?

PR #690, opened 2024-12-06
Fail the deployment when got the version conflict of Nucleus-Lite component

felicityzhao9:
*Description of changes:* If a customer deploys an aws.greengrass.NucleusLite component version that’s different from the version macro, we fail the deployment with an appropriate error message.

### DIFF
--- a/ggdeploymentd/src/deployment_handler.c
+++ b/ggdeploymentd/src/deployment_handler.c
@@ -36,6 +36,7 @@
 #include <ggl/uri.h>
 #include <ggl/utils.h>
 #include <ggl/vector.h>
+#include <ggl/version.h>
 #include <ggl/zip.h>
 #include <limits.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


ggdeploymentd/src/deployment_handler.c
```diff
@@ -1155,6 +1156,16 @@ static GglError resolve_dependencies(
             component_version = val->buf;
         }
 
+        if (ggl_buffer_eq(pair->key, GGL_STR("aws.greengrass.NucleusLite"))) {
+            if (!ggl_buffer_eq(component_version, GGL_STR(GGL_VERSION))) {
+                GGL_LOGE("The version of the aws.greengrass.NucleusLite "
+                         "component does not "
+                         "match the already installed version. "
```
patrzhan, line 1163: I think this should specify the version of the running GG Lite, since saying it doesn't match the installed version would make it sound like the software is deployed similar to how classic is which isn't true

patrzhan (approved):

PR #692, opened 2024-12-06
Add bootstrap functionality

sameerzuberi:
*Issue #, if available:*

*Description of changes:*

- When a bootstrap section is found in a component recipe during deployment we will save the deployment's status up to that point in the config. This includes the components that have already been deployed, the deployment type, and the deployment doc. 
- When `ggdeploymentd` starts up, it will first check the config to see if we have saved a previous in progress deployment. If one is found we will continue the deployment and complete it before returning to polling the deployment queue.
- When a bootstrap deployment is completed we delete the deployment status from the config. This ensures that we don't pick up the same deployment multiple times, or encounter conflicts when saving new deployments.
- Created a new library `core-bus-gghealthd` to provide a wrapper for retrieving a component's health status. This is used by both `gg-fleet-statusd` and `ggdeploymentd`.
- Updated `updated_component_name_vec` in `ggdeploymentd` to `components_to_deploy`, which now contains the name and version of each component that must be processed in the deployment. This allows us to save the component version to config as well in case of bootstrap.


ggdeploymentd/src/deployment_handler.c
```diff
@@ -2674,14 +2776,63 @@ static void handle_deployment(
...
+    } else {
+        GGL_LOGD(
+            "Found previously in progress deployment %.*s. Resuming "
+            "deployment.",
+            (int) bootstrap_deployment.deployment_id.len,
+            bootstrap_deployment.deployment_id.data
+        );
+        update_current_jobs_deployment(
+            bootstrap_deployment.deployment_id, GGL_STR("IN_PROGRESS")
+        );
```
cookpate, line 2807: IoT Jobs Listener may not have the Jobs ID for the deployment ID to update its state on-boot.

thomas-roos (approved):

ggdeploymentd/src/deployment_handler.c
```diff
@@ -2365,187 +2427,112 @@ static void handle_deployment(
...
+                GGL_LOGD(
+                    "Added %.*s to list of components that need to be "
+                    "processed.",
+                    (int) pair->key.len,
+                    pair->key.data
+                );
+            }
+
+            // Skip redeploying components in a RUNNING state
+            if (ggl_buffer_eq(component_status, GGL_STR("RUNNING"))) {
```
patrzhan, line 2483: Components in a FINISHED state should not be redeployed either

PR #693, opened 2024-12-06
Config reset json pointer support

patrzhan:
*Issue #, if available:*

*Description of changes:*
Moves json pointer parsing out from recipe runner to ggl-json for shared use.

Config reset supports json pointer keys now, instead of just single buffers (fixes this behavior).

Config reset supports an empty string resetting the whole configuration.

### DIFF
--- a/ggdeploymentd/src/component_config.c
+++ b/ggdeploymentd/src/component_config.c
@@ -6,12 +6,15 @@
 #include "deployment_model.h"
 #include <assert.h>
 #include <ggl/buffer.h>
+#include <ggl/constants.h>
 #include <ggl/core_bus/gg_config.h>
 #include <ggl/error.h>
+#include <ggl/json_pointer.h>
 #include <ggl/list.h>
 #include <ggl/log.h>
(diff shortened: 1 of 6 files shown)


archigup (approved):

ggdeploymentd/src/component_config.c
```diff
@@ -46,13 +49,46 @@ static GglError apply_reset_config(
...
+            ));
+            if (ret != GGL_ERR_OK) {
+                GGL_LOGE(
+                    "Error while deleting the component %.*s's configuration.",
+                    (int) component_name.len,
+                    component_name.data
+                );
+            }
+
+            break;
```
aws-kevinrickard, line 72: I think this should fail the deployment? Because if delete errors then either there was something wrong with their request they would want to know about, or something really bad happened in the config which they would probably also want to know about because the config state will not be as expected after that

aws-kevinrickard: Meant to put this on line 64

PR #702, opened 2024-12-09
Use config list for getting components list in FSS update; continue updating FSS even if one component fails

aws-kevinrickard:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/gg-fleet-statusd/src/fleet_status_service.c
+++ b/gg-fleet-statusd/src/fleet_status_service.c
@@ -4,6 +4,7 @@
 
 #include "fleet_status_service.h"
 #include <sys/types.h>
+#include <assert.h>
 #include <ggl/buffer.h>
 #include <ggl/bump_alloc.h>
 #include <ggl/cleanup.h>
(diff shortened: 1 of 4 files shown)


gg-fleet-statusd/src/fleet_status_service.c
```diff
@@ -167,36 +167,44 @@ GglError publish_fleet_status_update(
...
+            component->buf, &component_health
+        );
         if (ret != GGL_ERR_OK) {
             // skip reporting for components we fail to retrieve health statuses
-            // for
+            // for.
+            // TODO: Should all failures in this loop cause an error or warning
+            // and continue instead of return? because we want to update the
+            // status of any components we can, without one component affecting
+            // others
```
sameerzuberi, line 202: Agreed we should still publish updates for components that don't fail. Can that be included in this pr? If not we can leave this and come back to it

aws-kevinrickard: Added

sameerzuberi (approved):

PR #713, opened 2024-12-10
Save IoT Jobs version for bootstrap deployments

sameerzuberi:
*Issue #, if available:*

*Description of changes:*
Bootstrap deployments were not reporting their updated status to the cloud correctly due to a version conflict in IoT Jobs. This pr fixes that issue by saving the IoT jobs version to the config and retrieving and using it in case of bootstrap.

### DIFF
--- a/ggdeploymentd/src/bootstrap_manager.c
+++ b/ggdeploymentd/src/bootstrap_manager.c
@@ -16,6 +16,7 @@
 #include <ggl/map.h>
 #include <ggl/object.h>
 #include <ggl/vector.h>
+#include <inttypes.h>
 #include <limits.h>
 #include <stdbool.h>
 #include <stdint.h>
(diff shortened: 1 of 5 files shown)


ggdeploymentd/src/bootstrap_manager.c
```diff
@@ -111,6 +111,26 @@ GglError save_iot_jobs_id(GglBuffer jobs_id) {
     return GGL_ERR_OK;
 }
 
+GglError save_iot_jobs_version(int64_t jobs_version) {
+    GGL_LOGD("Saving IoT Jobs version %li in case of bootstrap.", jobs_version);
```
cookpate, line 115: Use the fixed-width macros for printing int64_t (build fails with warning arm32)
```suggestion
    GGL_LOGD("Saving IoT Jobs version %" PRIi64 " in case of bootstrap.", jobs_version);
```

cookpate: add `#include <inttypes.h>` to the include list for the macro

cookpate (approved):

PR #725, opened 2024-12-11
Skip reprocessing component artifacts for bootstrap components

sameerzuberi:
*Issue #, if available:*

*Description of changes:*
Encountered a bug where bootstrap components would redownload any artifacts listed in the recipe after rebooting. We now check if a component has already completed its bootstrap steps, if it has then we will skip to the part of the deployment that will only process other lifecycle stages in the recipe.

Added more robust logging.


thomas-roos (approved):
Did what it should do.

PR #732, opened 2024-12-12
Component cleanup only deletes config if no more versions active

patrzhan:
… gone

*Issue #, if available:*

*Description of changes:*
Component cleanup was deleting the config as part of component cleanup. We only want to do this if there are no more versions of that component needed.

### DIFF
--- a/ggdeploymentd/src/stale_component.c
+++ b/ggdeploymentd/src/stale_component.c
@@ -185,23 +185,26 @@ static GglError delete_component(
         (int) version_number.len,
         version_number.data
     );
+    GglError ret;
 
     // Remove component from config as we use that as source of truth for active
     // running components
-    GglError ret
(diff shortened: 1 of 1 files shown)


ggdeploymentd/src/stale_component.c
```diff
@@ -188,26 +188,28 @@ static GglError delete_component(
...
-        (int) component_name.len,
-        component_name.data
-    );
 
     static uint8_t root_path_mem[PATH_MAX];
     memset(root_path_mem, 0, sizeof(root_path_mem));
     GglBuffer root_path_buffer = GGL_BUF(root_path_mem);
 
-    ret = ggl_gg_config_read_str(
+    GglError ret = ggl_gg_config_read_str(
```
AniruddhaKanhere, line 212: I suggest you increase the scope of the above variable `ret` and reuse it here.
A bit cleaner.

AniruddhaKanhere (approved):

PR #742, opened 2024-12-13
Allow Fleet provision to run together with iotcored

rawalexe:
*Issue #, if available:*

*Description of changes:*


docs/examples/supported_lifecyle_types/1.json
```diff
@@ -0,0 +1,53 @@
+{
+  // Notice: All Key's are case sensitive
```
patrzhan, line 2: I don't think json supports comments like this (github also highlights it in red)

fleet-provisioning/src/entry.c
```diff
@@ -6,23 +6,27 @@
...
 #include <openssl/types.h>
 #include <openssl/x509.h>
-#include <string.h>
 #include <stdint.h>
 #include <stdio.h>
 
 #define MAX_TEMPLATE_LEN 129
 #define MAX_ENDPOINT_LENGTH 129
 #define MAX_TEMPLATE_PARAM_LEN 4096
+#define MAX_PATH_LEN 4096
```
patrzhan, line 29: Use PATH_MAX?

rawalexe: wanted to use custom here on purpose

fleet-provisioning/src/entry.c
```diff
@@ -187,8 +191,98 @@ static GglError fetch_from_db(FleetProvArgs *args) {
     return GGL_ERR_OK;
 }
 
-GglError run_fleet_prov(FleetProvArgs *args) {
-    GglError ret = fetch_from_db(args);
+static GglError update_cred_access(void) {
+    char *args[] = { "chown", "-R", "ggcore:ggcore", "/ggcredentials/", NULL };
```
patrzhan, line 195: While ggcore is our default, it is configurable by customer with GGL_SYSTEMD_SYSTEM_USER. So this doesn't work on any install that has configured the user to be something different.

rawalexe: if they are building it from the scratch then they will need to update this section also as there is no way for me to pull in cmake  configurations

fleet-provisioning/src/entry.c
patrzhan: Why?

fleet-provisioning/src/entry.c
patrzhan: Sounds good, maybe we can bring this out to be a macro so if they do want to configure this it's not buried too deep in the code

rawalexe: Added a warning in the docs

patrzhan (approved):

PR #759, opened 2024-12-18
Add static link packages

thomas-roos:
As the “NVIDIA JETPACK” is based on Ubuntu 22.04. the approach is to use the packages that are available (without adding some backport package soures list) and just compile the packages from source that are missing statically.  Use that static libs to link gg-lite static, but also link against shared libs that are available in the correct version. To do this I did create a container especially for this use case. Also some CMmakeLists improvements needs to be done to enable that static linking for just those libs that are not available with Ubuntu 22.04.

*Issue #, if available:*

*Description of changes:*


thomas-roos:
tested it

misc/staticbuildtestcontainer/getty-override.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
+ExecStart=
```
rawalexe, line 2: Why do you have a blank ExecStart here?

thomas-roos: good question - copied from here ;)
https://github.com/aws-greengrass/aws-greengrass-lite/blob/c14fc1848c32084f971570d5e4a60ed632afad65/misc/container/getty-override.conf#L2

PR #793, opened 2025-01-22
fix(fleet-provisioning): CSR creation null check

aslafy-z:
Errors are not catched.

Since `req` is a pointer to the `X509_REQ *` variable, it should be dereferenced.

*Description of changes:*
Dereference `req` in the error check.


rawalexe:
Thank you for catching the bug. We can merge this in

rawalexe (approved):

PR #802, opened 2025-01-28
Remove undesirable usage of static

aws-kevinrickard:
*Issue #, if available:*

*Description of changes:*

Fixes an issue where the buffer is reused on subsequent calls without being re-initialized to zero, causing messages from a previous call to show up in newer call logs.

Not needed for the 2.0.1 release, but can optionally be included.

### DIFF
--- a/ggdeploymentd/src/deployment_handler.c
+++ b/ggdeploymentd/src/deployment_handler.c
@@ -1278,7 +1278,7 @@ static GglError resolve_dependencies(
     }
 
     // Get list of thing groups
-    static uint8_t list_thing_groups_response_buf[2048] = { 0 };
+    uint8_t list_thing_groups_response_buf[2048] = { 0 };
     GglBuffer list_thing_groups_response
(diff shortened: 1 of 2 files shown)


ggdeploymentd/src/deployment_handler.c
```diff
@@ -1577,7 +1577,11 @@ static GglError resolve_dependencies(
 
         if (!found_local_candidate) {
             // Resolve with cloud and download recipe
-            static uint8_t resolve_component_candidates_response_buf[16384]
+            static uint8_t resolve_component_candidates_response_buf
+                [16384] // TODO: Clean up usages of static variables in this
+                        // file, many may be more appropriately non-static to
+                        // prevent issues where the memory is not re-initalized
+                        // to zero when the function is re-called
```
archigup, line 1584: If memory needs to be zeroed should be memset.
Though this might be fine as non-static, yeah.

PR #811, opened 2025-02-04
Enable TES server with each boot as well

rawalexe:
*Issue #, if available:*

*Description of changes:*


archigup:
Should this be enabled? Don't we only want to start it if a component needs it?

n9wxu (approved):

rawalexe:
it will start with the greengrass-lite.target. Env vars are only vended if mentioned in the recipe

PR #820, opened 2025-03-26
Switch the s3 download to using sigv4 library

AniruddhaKanhere:
*Issue #, if available:*

*Description of changes:*
This PR switches the S3 artifact downloader from curl sigv4 signing mechanism to using the aws-sigv4 library.
It provides 3 APIs:
- `aws_sigv4_add_header_for_signing` to allow users to add header key:value pairs to a given vector
- `aws_sigv4_s3_get_create_header` to generate the Authorization header which can be used with cURL
- `aws_sigv4_get_iso8601_time` to allow users to fetch the time in ISO 8601 format which has to be added to the headers and is also used by the SigV4 signing library.


modules/ggl-http/src/aws_sigv4.h
```diff
@@ -0,0 +1,62 @@
...
+
+#ifndef GGHTTPLIB_AWS_SIGV4_H
+#define GGHTTPLIB_AWS_SIGV4_H
+
+#include "ggl/http.h"
+#include <ggl/buffer.h>
+#include <ggl/error.h>
+#include <ggl/vector.h>
+
+#define ZERO_PAYLOAD_SHA \
```
archigup, line 13: Could you put a comment on how to get this value? I.e. sha-256 hash of ""

modules/ggl-http/src/aws_sigv4.h
```diff
@@ -0,0 +1,62 @@
...
+
+#define HTTP_METHOD_GET "GET" /**< HTTP Method GET string. */
+#define HTTP_METHOD_PUT "PUT" /**< HTTP Method PUT string. */
+#define HTTP_METHOD_POST "POST" /**< HTTP Method POST string. */
+#define HTTP_METHOD_HEAD "HEAD" /**< HTTP Method HEAD string. */
+
+#define X_AMZ_SECURITY_TOKEN "x-amz-security-token"
+#define X_AMZ_DATE "x-amz-date"
+#define X_AMZ_CONTENT_SHA256 "x-amz-content-sha256"
+#define HOST_HEADER "host"
```
archigup, line 24: These macros don't seem very useful. They are constants with meaning that the macro name doesnt add to. Would be better to just use the string literals

modules/ggl-http/src/aws_sigv4.c
```diff
@@ -0,0 +1,348 @@
...
+}
+
+GglError aws_sigv4_add_header_for_signing(
+    GglByteVec *vector, GglBuffer header_key, GglBuffer header_value
+) {
+    GglError ret = ggl_byte_vec_append(vector, header_key);
+    ggl_byte_vec_chain_push(&ret, vector, ':');
+    ggl_byte_vec_chain_append(&ret, vector, header_value);
+
+    // Non-canonical delimiter used by the AWS Sigv4 library to separate
```
archigup, line 90: What does non-cannonical mean here? isn't \r\n the normal HTTP separator between headers in a request?

AniruddhaKanhere: I think `\n` is the delimiter used by Sigv4 library when calculating the hash. Thus, this is non-canonical from that perspective.

modules/ggl-http/src/aws_sigv4.h
AniruddhaKanhere: Yup.

modules/ggl-http/src/aws_sigv4.c
```diff
@@ -0,0 +1,348 @@
...
+    ggl_byte_vec_chain_append(&ret, vector, GGL_STR("\r\n"));
+
+    return ret;
+}
+
+static GglError translate_sigv4_error(SigV4Status_t status) {
+    GglError ret;
+
+    switch (status) {
+    /**
```
archigup, line 101: These shouldnt be doc comments (not attached to a fn/var/etc)

modules/ggl-http/src/aws_sigv4.h
```diff
@@ -0,0 +1,62 @@
...
+    GglBuffer host;
+} S3RequiredHeaders;
+
+/// Add HTTP headers as key:value pairs to a given vector so that it can be
+/// later on signed and added to the SigV4 header.
+GglError aws_sigv4_add_header_for_signing(
+    GglByteVec *vector, GglBuffer header_key, GglBuffer header_value
+);
+
+size_t aws_sigv4_get_iso8601_time(char buffer[17]);
```
cookpate, line 52: There is no mechanism to guarantee a C array will be at least 17 bytes passed into a function like this. Maybe move the array into a struct i.e.
```c
typedef struct Sigv4Timestamp {
    uint8_t bytes[17]; // YYYYMMDDTHHMMSSz\0
} Sigv4Timestamp;
```

cookpate: Nevermind, this isn't part of the public API

AniruddhaKanhere: I changed the signature of the function.

modules/ggl-http/src/aws_sigv4.c
```diff
@@ -0,0 +1,348 @@
...
+    struct timeval tv;
+    struct tm *tm_info;
+
+    gettimeofday(&tv, NULL);
+
+    // NOLINTNEXTLINE(concurrency-mt-unsafe)
+    tm_info = gmtime(&tv.tv_sec);
+
+    return strftime(buffer, 17, "%Y%m%dT%H%M%SZ", tm_info);
+}
```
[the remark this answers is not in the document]
AniruddhaKanhere: Good catch. Done.

modules/aws-sigv4/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+ggl_init_module(
+  aws_sigv4
+  SRCDIR ${aws_sigv4_SOURCE_DIR}/source
+  INCDIRS config ${aws_sigv4_SOURCE_DIR}/source/include)
+
+target_compile_definitions(aws_sigv4 PRIVATE AWS_SIGV4_SOURCE)
+target_compile_options(aws_sigv4 PRIVATE "-Wno-conversion")
```
archigup, line 11: Can you add a comment to remove this when updating sigv4 library?

modules/aws-sigv4/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# aws-greengrass-lite - AWS IoT Greengrass runtime for constrained devices
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+ggl_init_module(
+  aws_sigv4
+  SRCDIR ${aws_sigv4_SOURCE_DIR}/source
+  INCDIRS config ${aws_sigv4_SOURCE_DIR}/source/include)
+
+target_compile_definitions(aws_sigv4 PRIVATE AWS_SIGV4_SOURCE)
```
archigup, line 10: Why is this needed?

archigup (approved):

PR #821, opened 2025-03-27
Add HTTP proxy support to iotcored

cookpate:
Allows MQTT connections on networks where all outbound communication is gated by a proxy server.

Tested by running a deployment to a remote tinyproxy server. Configure the network proxy by writing the following configuration object to services/aws.greengrass.NucleusLite/configuration/networkProxy:
```json
"networkProxy": {
    "noProxyAddresses": "192.168.1.1,*.example.com:80",
    "proxy": {
        "url": "http://www.myproxy.io:8888"
    }
}
```
These settings are equivalent to setting no_proxy and https_proxy environment variables for OpenSSL

*Description of changes:*
* Split TLS connect flow into different paths for http proxy and no proxy.
* Determine based on environment variables and configuration database whether to connect with a proxy

### DIFF
--- a/bins/iotcored/src/entry.c
+++ b/bins/iotcored/src/entry.c
@@ -8,13 +8,101 @@
 #include <ggl/buffer.h>
 #include <ggl/core_bus/gg_config.h>
 #include <ggl/error.h>
+#include <ggl/log.h>
 #include <ggl/object.h>
 #include <limits.h>
 #include <string.h>
+#include <stdbool.h>
 #include <stdint.h>
+#include <stdlib.h>
 
 #define MAX_ENDPOINT_LEN 128
 #define MAX_THINGNAME_LEN 128
+
+static bool get_proxy_variable(GglBufList aliases, GglBuffer *destination) {
+    for (size_t i = 0; i < aliases.len; ++i) {
+        char *name = (char *) aliases.bufs[i].data;
+        if (name == NULL) {
+            continue;
+        }
+        // This is safe as long as getenv is reentrant
+        // and no other threads call setenv.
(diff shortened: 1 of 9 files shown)


bins/iotcored/bin/iotcored.c
```diff
@@ -57,6 +59,20 @@ static struct argp argp = { opts, arg_parser, 0, doc, 0, 0, 0 };
 int main(int argc, char **argv) {
     static IotcoredArgs args = { 0 };
 
+    // NOLINTBEGIN(concurrency-mt-unsafe)
```
archigup, line 62: Could you leave a comment mentioning this is safe as long as getenv is re-entrant on the target and setenv is not called by other threads?

cookpate: Oops, I think I have two different spots where getenv is called... Should probably be in entry.c where the storage for the loaded config values is located.

archigup (dismissed):

bins/iotcored/src/entry.c
```diff
@@ -8,14 +8,102 @@
...
+    if (args->no_proxy == NULL) {
+        GglBuffer no_proxy = GGL_BUF(no_proxy_mem);
+        no_proxy.len -= 1;
+        GglError ret = ggl_gg_config_read_str(
+            GGL_BUF_LIST(
+                GGL_STR("services"),
+                GGL_STR("aws.greengrass.NucleusLite"),
+                GGL_STR("configuration"),
+                GGL_STR("networkProxy"),
+                GGL_STR("noproxy"),
```
cookpate, line 97:
```suggestion
                GGL_STR("noProxyAddresses"),
```

archigup (approved):

PR #835, opened 2025-04-08
Print out HTTP error codes

rawalexe:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/modules/ggdeploymentd/src/deployment_handler.c
+++ b/modules/ggdeploymentd/src/deployment_handler.c
@@ -12,8 +12,10 @@
 #include "stale_component.h"
 #include <sys/types.h>
 #include <assert.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <ggl/alloc.h>
+#include <ggl/backoff.h>
 #include <ggl/base64.h>
 #include <ggl/buffer.h>
 #include <ggl/bump_alloc.h>
@@ -43,6 +45,7 @@
 #include <ggl/zip.h>
 #include <limits.h>
 #include <string.h>
+#include <unistd.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdlib.h>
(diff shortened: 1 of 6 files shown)


modules/ggdeploymentd/src/deployment_handler.c
```diff
@@ -409,6 +410,84 @@ static GglError get_tes_credentials(TesCredentials *tes_creds) {
     return GGL_ERR_OK;
 }
 
+typedef struct CurlRequestRetryCtx {
```
archigup, line 413: Why does this have a struct tag?

test_modules/s3-get-test/src/entry.c
```diff
@@ -112,20 +112,21 @@ GglError run_s3_test(char *region, char *bucket, char *key, char *file_path) {
...
-                .session_token = aws_session_token->buf,
-            }
+            (SigV4Details) { .aws_region = ggl_buffer_from_null_term(region),
+                             .aws_service = GGL_STR("s3"),
+                             .access_key_id = aws_access_key_id->buf,
+                             .secret_access_key = aws_secret_access_key->buf,
+                             .session_token = aws_session_token->buf },
+            &http_response_code
         );
+        (void) http_response_code;
```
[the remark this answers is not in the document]
rawalexe: clang was giving me a hard time.

modules/ggdeploymentd/src/deployment_handler.c
rawalexe: Oops

modules/ggl-http/include/ggl/http.h
```diff
@@ -98,7 +98,8 @@ GglError sigv4_download(
     GglBuffer host,
     GglBuffer file_path,
     int fd,
-    SigV4Details sigv4_details
+    SigV4Details sigv4_details,
+    long *http_response_code
```
archigup, line 102: This should probably be a uint32_t or uint64_t, etc, depending on range of value.

modules/ggl-file/include/ggl/file.h
```diff
@@ -87,4 +87,6 @@ static inline void cleanup_closedir(DIR **dirp) {
     }
 }
 
+GglError truncate_file(void *response_data);
```
archigup, line 90: This needs to be prefixed

archigup: This also is not atomic so should be used with care

archigup: Does this really need to be a general use function, and not a private helper?

modules/ggdeploymentd/src/deployment_handler.c
```diff
@@ -409,6 +410,84 @@ static GglError get_tes_credentials(TesCredentials *tes_creds) {
...
+        &http_response_code
+    );
+    if (http_response_code == (long) 403) {
+        GglError err = retry_ctx->retry_cleanup_fn(retry_ctx->response_data);
+        GGL_LOGE(
+            "Artifact download attempt failed with 403. Retrying with backoff."
+        );
+        if (err != GGL_ERR_OK) {
+            retry_ctx->err = err;
+            return GGL_ERR_OK;
```
AniruddhaKanhere, line 446: Why do we return success when there has been a failure? Should that not be only for happy path?

AniruddhaKanhere (changes requested):

modules/ggdeploymentd/src/deployment_handler.c
```diff
@@ -409,6 +410,84 @@ static GglError get_tes_credentials(TesCredentials *tes_creds) {
...
+        GGL_LOGE(
+            "Artifact download attempt failed due to error: %d", ret
+
+        );
+        retry_ctx->err = ret;
+        return GGL_ERR_OK;
+    }
+
+    retry_ctx->err = ret;
+    return GGL_ERR_OK;
```
AniruddhaKanhere, line 460: We return OK on line 456 too? It is confusing as to why we do that.

modules/ggl-file/include/ggl/file.h
AniruddhaKanhere: If this has to be public, then we should add a lot of safeguards against unintended use. Otherwise make it private

modules/ggl-file/src/file.c
```diff
@@ -578,3 +578,18 @@ GglError ggl_file_read_path(GglBuffer path, GglBuffer *content) {
...
     return GGL_ERR_OK;
 }
+
+GglError truncate_file(void *response_data) {
+    int fd = *(int *) response_data;
+
+    int ret;
+    do {
+        ret = ftruncate(fd, 0);
+    } while ((ret == -1) && (errno == EINTR));
```
AniruddhaKanhere, line 588: This errno might be overwritten by something else if used incorrectly. Not thread safe

modules/ggl-http/include/ggl/http.h
AniruddhaKanhere: I think even a 16 bit would be enough no. But if we are being realistic with padding etc involved, then it can be 32 bit too. No need for a long.

rawalexe: it needs to be long or equivalent according to the doc: https://curl.se/libcurl/c/CURLINFO_RESPONSE_CODE.html#:~:text=Pass%20a%20pointer%20to%20a%20long%20to%20receive%20the%20last%20received%20HTTP%2C%20FTP%2C%20SMTP%20or%20LDAP%20(OpenLDAP%20only)%20response%20code.%20This%20option%20was%20previously%20known%20as%20CURLINFO_HTTP_CODE%20in%20libcurl%207.10.7%20and%20earlier.%20The%20stored%20value%20is%20zero%20if%20no%20server%20response%20code%20has%20been%20received.

modules/ggl-file/include/ggl/file.h
rawalexe: I needed two instances of this, one in http lib another in ggdeploymentd, So, I'll add more safe guards to it

modules/ggl-http/include/ggl/http.h
rawalexe: will use int32_t`

modules/ggdeploymentd/src/deployment_handler.c
rawalexe: Ok to stop the retry

AniruddhaKanhere (approved):

AniruddhaKanhere (approved):

PR #837, opened 2025-04-10
Update IPC clients to understand IPC errors

AniruddhaKanhere:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/modules/eventstream/src/rpc.c
+++ b/modules/eventstream/src/rpc.c
@@ -14,7 +14,7 @@
 GglError eventstream_get_common_headers(
     EventStreamMessage *msg, EventStreamCommonHeaders *out
 ) {
-    int32_t message_type = -1;
+    int32_t message_type = 0;
     int32_t message_flags = 0;
     int32_t stream_id = 0;
 
@@ -27,18 +27,21 @@ GglError eventstream_get_common_headers(
                 GGL_LOGE(":message-type header not Int32.");
                 return GGL_ERR_INVALID;
             }
+
             message_type = header.value.int32;
         } else if (ggl_buffer_eq(header.name, GGL_STR(":message-flags"))) {
             if (header.value.type != EVENTSTREAM_INT32) {
                 GGL_LOGE(":message-flags header not Int32.");
                 return GGL_ERR_INVALID;
             }
+
             message_flags = header.value.int32;
         } else if (ggl_buffer_eq(header.name, GGL_STR(":stream-id"))) {
             if (header.value.type != EVENTSTREAM_INT32) {
                 GGL_LOGE(":stream-id header not Int32.");
                 return GGL_ERR_INVALID;
             }
+
             stream_id = header.value.int32;
         }
     }
@@ -46,5 +49,6 @@ GglError eventstream_get_common_headers(
     out->message_type = message_type;
     out->message_flags = message_flags;
     out->stream_id = stream_id;
+
     return GGL_ERR_OK;
 }
--- a/modules/ggipc-client/src/client.c
+++ b/modules/ggipc-client/src/client.c
@@ -5,6 +5,7 @@
 #include "ggipc/client.h"
 #include <sys/types.h>
 #include <assert.h>
+#include <ggl/alloc.h>
 #include <ggl/base64.h>
 #include <ggl/buffer.h>
 #include <ggl/bump_alloc.h>
@@ -17,6 +18,7 @@
 #include <ggl/eventstream/types.h>
 #include <ggl/file.h>
 #include <ggl/io.h>
+#include <ggl/ipc/common.h>
 #include <ggl/json_decode.h>
 #include <ggl/json_encode.h>
 #include <ggl/log.h>
@@ -203,9 +205,11 @@ GglError ggipc_connect_auth(GglBuffer socket_path, GglBuffer *svcuid, int *fd) {
 GglError ggipc_call(
     int conn,
     GglBuffer operation,
+    GglBuffer service_model_type,
     GglMap params,
     GglAlloc *alloc,
-    GglObject *result
+    GglObject *result,
+    GglIpcError *remote_err
 ) {
     EventStreamHeader headers[] = {
         { GGL_STR(":message-type"),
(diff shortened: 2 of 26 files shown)


modules/eventstream/src/rpc.c
```diff
@@ -27,24 +32,58 @@ GglError eventstream_get_common_headers(
                 GGL_LOGE(":message-type header not Int32.");
                 return GGL_ERR_INVALID;
             }
+            if (message_type != -1) {
+                GGL_LOGE("Received duplicate header ':message-type'.");
+                return GGL_ERR_INVALID;
+            }
```
archigup, line 38: Current SDKs/Classic function with duplicate headers (they just use last value).

modules/eventstream/src/rpc.c
```diff
@@ -9,14 +9,19 @@
...
+// TODO: Refactor this function
+//  NOLINTNEXTLINE(readability-function-cognitive-complexity)
 GglError eventstream_get_common_headers(
     EventStreamMessage *msg, EventStreamCommonHeaders *out
 ) {
     int32_t message_type = -1;
-    int32_t message_flags = 0;
-    int32_t stream_id = 0;
+    int32_t message_flags = -1;
+    int32_t stream_id = -1;
```
archigup, line 22: These will then be in an invalid state if the header is missing.

modules/ggipc-client/src/client.c
```diff
@@ -157,6 +184,15 @@ GglError ggipc_connect_auth(GglBuffer socket_path, GglBuffer *svcuid, int *fd) {
         return GGL_ERR_FAILURE;
     }
 
+    if (common_headers.content_type.data != NULL) {
```
archigup, line 187: We don't use conack payload so content_type doesn't matter (and would be unexpected if present)

modules/ggipc-client/src/client.c
```diff
@@ -239,23 +279,13 @@ GglError ggipc_call(
...
-    }
-
-    if (common_headers.message_type != EVENTSTREAM_APPLICATION_MESSAGE) {
-        GGL_LOGE(
-            "Stream %" PRId32 " received invalid message type: %" PRId32 ".",
-            common_headers.stream_id,
-            common_headers.message_type
-        );
-        return GGL_ERR_FAILURE;
+    if (common_headers.content_type.data != NULL) {
```
archigup, line 282: We always expect a payload for application message responses, so this should be required.

modules/ggipc-client/src/client.c
```diff
@@ -272,6 +302,12 @@ GglError ggipc_call(
         }
     }
 
+    if (common_headers.message_type == EVENTSTREAM_APPLICATION_ERROR) {
+        *remote_err = GGL_ERR_FAILURE;
```
archigup, line 306: remote_err should be set according to IPC err response

modules/ggipc-client/src/client.c
```diff
@@ -239,23 +279,13 @@ GglError ggipc_call(
...
-        GGL_LOGI(
-            "Received IPC error on stream %" PRId32 ". Payload: %.*s.",
-            common_headers.stream_id,
-            (int) msg.payload.len,
-            msg.payload.data
-        );
-        return GGL_ERR_FAILURE;
-    }
-
-    if (common_headers.message_type != EVENTSTREAM_APPLICATION_MESSAGE) {
```
archigup, line 252: We still need to ensure that we get an APPLICATION_MESSAGE or APPLICATION_ERROR

modules/ggipc-client/src/client.c
archigup: Maybe it should be an IPC error code enum instead of GglError?

archigup: We can always parse the error response here, log the code and message, and return the code enum value

archigup: Another option would be to alloc the message in the passed-in allocator and return the  IpcError type

modules/ggipc-client/src/client.c
```diff
@@ -463,7 +573,41 @@ GglError ggipc_publish_to_iot_core(
...
+        GglBuffer message;
+        ret = parse_server_error(resp, &error, &message);
+        if (ret != GGL_ERR_OK) {
+            GGL_LOGE("Failed to parse server response.");
+            return GGL_ERR_FAILURE;
+        }
+
+        if (ggl_buffer_eq(error, GGL_STR("aws.greengrass#UnauthorizedError"))) {
+            GGL_LOGE(
+                "Componenet unauthorized: %.*s", (int) message.len, message.data
```
archigup, line 603:
```suggestion
                "Component unauthorized: %.*s", (int) message.len, message.data
```

modules/ggipc-client/src/client.c
```diff
@@ -463,7 +573,41 @@ GglError ggipc_publish_to_iot_core(
...
+        alloc,
+        &resp,
+        &remote_error
     );
+
+    if (ret != GGL_ERR_OK) {
+        return ret;
+    }
+
+    if (remote_error != GGL_ERR_OK) {
```
archigup, line 592: We should only use value of remote_error if return val is GGL_ERR_REMOTE

modules/ggipc-client/src/client.c
```diff
@@ -272,6 +302,12 @@ GglError ggipc_call(
         }
     }
 
+    if (common_headers.message_type == EVENTSTREAM_APPLICATION_ERROR) {
+        *remote_err = GGL_ERR_FAILURE;
+    } else {
+        *remote_err = GGL_ERR_OK;
```
archigup, line 308: Might be better to

modules/ggipc-client/src/client.c
archigup, line 306: Should return GGL_ERR_REMOTE when there is a remote error

archigup: Though here we need to translate instead of having the wrapper propagate the error code

modules/ggipc-client/src/client.c
```diff
@@ -240,20 +246,63 @@ GglError ggipc_call(
...
+
+        ret = ggl_map_validate(
+            err_result.map,
+            GGL_MAP_SCHEMA(
+                { GGL_STR("_errorCode"), true, GGL_TYPE_BUF, &error_code_obj },
+                { GGL_STR("_message"), false, GGL_TYPE_BUF, &message_obj },
+            )
+        );
+        if (ret != GGL_ERR_OK) {
+            GGL_LOGE("Error response does not match known schema.");
```
archigup, line 282:
```suggestion
            GGL_LOGE("Error response does not match known schema.");
            return ret;
```

modules/eventstream/src/rpc.c
```diff
@@ -14,7 +14,7 @@
 GglError eventstream_get_common_headers(
     EventStreamMessage *msg, EventStreamCommonHeaders *out
 ) {
-    int32_t message_type = -1;
+    int32_t message_type = 0;
```
archigup, line 17: Doesn't this mean we now default to APPLICATION_MESSAGE when we dont get a header?
Probably fine to do that.

archigup (approved):

PR #848, opened 2025-04-17
CMake, libcurl: Lower version requirements

thomas-roos:
As sig4 signing is now not done with libcurl anymore this requirement can be removed. Also CMake version is set to 3.22, which is the version that is provided by kirkstone yocto lts version.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,7 +253,7 @@ endforeach()
 find_package(PkgConfig REQUIRED)
 
 pkg_search_module(openssl REQUIRED IMPORTED_TARGET openssl)
-pkg_search_module(libcurl REQUIRED IMPORTED_TARGET libcurl)
+pkg_search_module(libcurl REQUIRED IMPORTED_TARGET libcurl>=7.82)
 pkg_search_module(sqlite3 REQUIRED IMPORTED_TARGET sqlite3)
 pkg_search_module(yaml REQUIRED IMPORTED_TARGET yaml-0.1)
 pkg_search_module(libsystemd REQUIRED IMPORTED_TARGET libsystemd)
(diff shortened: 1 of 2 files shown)


cookpate:
We will want a version requirement, but we are not sure what the minimum version is.

thomas-roos:
I would be okay with 7.82 - that is included in yocto kirkstone (https://layers.openembedded.org/layerindex/recipe/300818/)

CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-cmake_minimum_required(VERSION 3.25)
```
archigup, line 5: This will need some investigation; cmake versions also change the semantics of functions, so we need to make sure that this isn't silently changing the build output in some way.

archigup: Should be fine; we were previously also using some of the SYSTEM stuff added in 3.25, but seems not currently used.

thomas-roos: run UAT is not enough?
I know it build and connect to the cloud with the proposed changes.
What should I look for?

cookpate:
What's the lowest version of libcurl you have used to build a lite image with? I seem unable to bootstrap OpenSSL 3.0.x on Ubuntu 20.04 to test with its curl 7.68.

thomas-roos:
I think it was 7.82 - that is included in yocto kirkstone

archigup:
Lets set it to 7.82 then

thomas-roos:
done, please approve

cookpate (approved):

PR #852, opened 2025-04-22
Add error handling and error detail to yaml parser

rawalexe:
*Issue #, if available:*

*Description of changes:*


modules/ggl-yaml/src/yaml_decode.c
```diff
@@ -203,12 +203,33 @@ GglError ggl_yaml_decode_destructive(
...
+    if (!yaml_parser_load(&parser, &document)) {
+        GGL_LOGE(
+            "Yaml parser load failed. Parser error: %s, at line %zu, column "
+            "%zu",
+            parser.problem,
+            parser.problem_mark.line + 1,
+            parser.problem_mark.column + 1
+        );
+        yaml_parser_delete(&parser);
+        return GGL_ERR_FATAL;
```
aws-kevinrickard, line 226: Why Fatal in this case? Seems like this is recoverable

rawalexe: well the document isn't valid, How can I recover?

modules/ggl-yaml/src/yaml_decode.c
```diff
@@ -203,12 +203,33 @@ GglError ggl_yaml_decode_destructive(
...
-    yaml_parser_load(&parser, &document);
+    if (!yaml_parser_load(&parser, &document)) {
+        GGL_LOGE(
+            "Yaml parser load failed. Parser error: %s, at line %zu, column "
+            "%zu",
+            parser.problem,
+            parser.problem_mark.line + 1,
+            parser.problem_mark.column + 1
+        );
+        yaml_parser_delete(&parser);
```
AniruddhaKanhere, line 225: Does this ever fail?

modules/ggl-yaml/src/yaml_decode.c
```diff
@@ -203,12 +203,33 @@ GglError ggl_yaml_decode_destructive(
...
+        return GGL_ERR_PARSE;
+    }
     yaml_node_t *root_node = yaml_document_get_root_node(&document);
+    if (root_node == NULL) {
+        GGL_LOGE("Yaml document is empty.");
+        return GGL_ERR_PARSE;
```
AniruddhaKanhere, line 231: Why is this an error. Why can a YAML not be empty?

modules/ggl-yaml/src/yaml_decode.c
AniruddhaKanhere: I see that it returns void so should never fail.

rawalexe: In a rare occurrence it may, the parser it self contains that information but it's similar to failing a free

AniruddhaKanhere: https://github.com/yaml/libyaml/blob/master/include/yaml.h#L1332.

modules/ggl-yaml/src/yaml_decode.c
rawalexe: We process a recipe with certain values for a Generic component, a generic component with no recipe does not makes sense.

modules/ggl-yaml/src/yaml_decode.c
```diff
@@ -203,12 +203,33 @@ GglError ggl_yaml_decode_destructive(
     static pthread_mutex_t mtx = PTHREAD_MUTEX_INITIALIZER;
     GGL_MTX_SCOPE_GUARD(&mtx);
 
+    GGL_LOGT(
+        "%s received yaml content: %.*s", __func__, (int) buf.len, buf.data
+    );
+
     static yaml_parser_t parser;
-    yaml_parser_initialize(&parser);
+    if (!yaml_parser_initialize(&parser)) {
```
AniruddhaKanhere, line 211: As a todo, we should figure out a way to not have a parser which calls malloc.

modules/ggl-yaml/src/yaml_decode.c
AniruddhaKanhere: Umm, but that should be decided by the caller.

YAML wrappers should not make an assumption on what is right and what is not.

archigup: This should be error
Empty file is not a valid yaml file

archigup: Yaml file must contain at least one yaml document

modules/ggl-yaml/src/yaml_decode.c
archigup: I'm not crazy enough to write our own YAML parser

Though could be an option if we restrict to a subset of YAML.

PR #853, opened 2025-04-22
Modify yaml lib to return NOENTRY on empty file

rawalexe:
*Issue #, if available:*

*Description of changes:*


patrzhan (approved):

modules/ggl-yaml/src/yaml_decode.c
```diff
@@ -223,7 +223,7 @@ GglError ggl_yaml_decode_destructive(
             parser.problem_mark.column + 1
         );
         yaml_parser_delete(&parser);
-        return GGL_ERR_PARSE;
+        return GGL_ERR_NOENTRY;
```
AniruddhaKanhere, line 226: I think that is supposed to be on line 231.

PR #854, opened 2025-04-22
Update alloc/arena interfaces

archigup:


modules/core-bus-gghealthd/src/gg_healthd.c
```diff
@@ -3,44 +3,43 @@
...
 #include <stdint.h>
 
 GglError ggl_gghealthd_retrieve_component_status(
-    GglBuffer component, GglBuffer *component_status
+    GglBuffer component, GglArena *alloc, GglBuffer *component_status
 ) {
-    static uint8_t buffer[10 * sizeof(GglObject)] = { 0 };
-    GglBumpAlloc balloc = ggl_bump_alloc_init(GGL_BUF(buffer));
+    static uint8_t resp_mem[10 * sizeof(GglObject)] = { 0 };
+    GglArena resp_alloc = ggl_arena_init(GGL_BUF(resp_mem));
```
AniruddhaKanhere, line 19: Why not pass in the parameter from the function signature instead of creating a new allocator here?

AniruddhaKanhere (changes requested):
As is the custom, leaving this here so that this PR doesn't get merged before I can finish reviewing this. The question is not really a blocker :)

modules/core-bus-gghealthd/src/gg_healthd.c
archigup: This memory should not consume space in callers allocator. The parameter is for allocating things returned to the caller. This stuff is used in this function and not returned to caller so should not be using caller's memory after returning.

AniruddhaKanhere (approved):

modules/core-bus-gghealthd/src/gg_healthd.c
```diff
@@ -3,44 +3,43 @@
...
         &result
     );
-    if (call_error != GGL_ERR_OK) {
-        return call_error;
-    }
-    if (method_error != GGL_ERR_OK) {
-        return method_error;
+    if (ret != GGL_ERR_OK) {
+        if (ret == GGL_ERR_REMOTE) {
+            return method_error;
```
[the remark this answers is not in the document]
archigup: Yeah, might make sense to log. The core bus helpers used to have an err out parameter, to check on remote error. But every call site just immediately used in in place of the error, so I had the helpers just do that.

PR #861, opened 2025-04-25
Make map and list headers public

archigup:
Adds the list and map headers to the public headers of the sdk.

Updates map validation to not use a bool for configuring whether fields are required.

Also has a commit improving type checking on GGL_STR.


modules/core-bus-aws-iot-mqtt/src/aws_iot_mqtt.c
```diff
@@ -84,8 +85,8 @@ GglError ggl_aws_iot_mqtt_subscribe_parse_resp(
     GglError ret = ggl_map_validate(
         response,
         GGL_MAP_SCHEMA(
-            { GGL_STR("topic"), true, GGL_TYPE_BUF, &topic_obj },
-            { GGL_STR("payload"), true, GGL_TYPE_BUF, &payload_obj },
+            { GGL_STR("topic"), GGL_REQUIRED, GGL_TYPE_BUF, &topic_obj },
+            { GGL_STR("payload"), GGL_REQUIRED, GGL_TYPE_BUF, &payload_obj },
```
rawalexe, line 89: Isn't it a bit excessive?

archigup: Nah, this makes it clear what the state is. Code can be reviewed without having to go look up which way the bool is.

AniruddhaKanhere (approved):

PR #872, opened 2025-05-08
Reduce size of GglObject

archigup:


modules/ggl-sdk/src/object.c
```diff
@@ -3,105 +3,171 @@
...
-    sizeof(GglObject) == sizeof(GglObjectPriv), "GglObject impl invalid."
+    sizeof(void *) == sizeof(size_t),
+    "Platforms where pointers and size_t are not same width are not currently "
+    "supported."
 );
+
 static_assert(
-    alignof(GglObject) == alignof(GglObjectPriv), "GglObject impl invalid."
+    (sizeof(size_t) == 4) || (sizeof(size_t) == 8),
+    "Platform not 32 or 64 bits?"
```
aws-kevinrickard, line 22: nit: I think this message should be more clear, e.g. Say only 32 or 64 bit systems are currently supported, if that's what we are verifying

aws-kevinrickard (dismissed):

modules/ggl-sdk/src/object.c
archigup: Its not entirely sure, since we're assuming pointers are word sized. This will also trip on CHERI, which we can add support for at some point if we have a customer.

modules/ggl-sdk/src/object.c
```diff
@@ -3,105 +3,171 @@
...
-    GglObjectPriv result;
-    memcpy(&result, &obj, sizeof(GglObject));
-    return result;
-}
-
 GglObjectType ggl_obj_type(GglObject obj) {
-    return (GglObjectType) priv_from_obj(obj).type;
+    // Last byte is tag
+    uint8_t result = obj._private[sizeof(obj._private) - 1];
+    assert(result <= 7);
```
cookpate, line 28: The `GglObjectType` enum is 0-indexed, so `GGL_TYPE_MAP` is 6.
```suggestion
    assert(result < 7);
```

archigup: ah

modules/ggl-sdk/src/object.c
aws-kevinrickard: I see. Either way is fine then

aws-kevinrickard (dismissed):

patrzhan (approved):

PR #873, opened 2025-05-09
Add library function for parsing docker URI

patrzhan:
*Issue #, if available:*

*Description of changes:*
Add a function to parse docker URI into its various components. Assume it is in the format `[registry-host][:port]/[username/]repository[:tag][@algo:digest]`. 

Manually tested the function against example docker URIs:

```
ubuntu:20.04

nginx@sha256:a7705264f3d1f61b30c65e06c3b8b4c32b89c770cdafe39927cbe21d62a4744a

public.ecr.aws/cloudwatch-agent/cloudwatch-agent:latest
```

### DIFF
--- a/modules/ggl-uri/src/gghttp_uri.c
+++ b/modules/ggl-uri/src/gghttp_uri.c
@@ -1,4 +1,5 @@
 #include "ggl/uri.h"
+#include <assert.h>
 #include <ggl/arena.h>
 #include <ggl/buffer.h>
 #include <ggl/error.h>
@@ -7,6 +8,7 @@
 #include <uriparser/Uri.h>
 #include <uriparser/UriBase.h>
 #include <stdalign.h>
+#include <stdbool.h>
 #include <stdint.h>
 
 static GglBuffer buffer_from_text_range(UriTextRangeA range) {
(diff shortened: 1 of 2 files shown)


modules/ggl-uri/src/gghttp_uri.c
```diff
@@ -136,3 +138,283 @@ GglError gg_uri_parse(GglArena *arena, GglBuffer uri, GglUriInfo *info) {
...
+static GglError find_docker_uri_separators(
+    GglBuffer uri,
+    size_t slashes[2],
+    size_t *slash_count,
+    size_t colons[3],
+    size_t *colon_count,
+    size_t *at,
+    bool *has_registry
+) {
+    for (size_t i = uri.len; i > 0; i--) {
```
rawalexe, line 151: meaningful variable name will be helpful

modules/ggl-uri/src/gghttp_uri.c
```diff
@@ -136,3 +138,283 @@ GglError gg_uri_parse(GglArena *arena, GglBuffer uri, GglUriInfo *info) {
 
     return GGL_ERR_OK;
 }
+
+static GglError find_docker_uri_separators(
+    GglBuffer uri,
+    size_t slashes[2],
```
AniruddhaKanhere, line 144: Why not get a pointer instead of an array? This doesn't help the compiler. If you pass in a pointer which points to only one valid location (instead of 2), compiler won't catch the bug.

modules/ggl-uri/src/gghttp_uri.c
AniruddhaKanhere, line 151: A minor suggestion: make this like so
```suggestion
    for (size_t len = uri.len - 1; len >= 0; len--) {
```

This will allow you to use `len` everywhere instead of `len - 1`.

AniruddhaKanhere: You'd need to check before though that `uri.len` is non 0. And return apt error code if that is the case.

modules/ggl-uri/src/gghttp_uri.c
```diff
@@ -136,3 +138,283 @@ GglError gg_uri_parse(GglArena *arena, GglBuffer uri, GglUriInfo *info) {
...
+                *colon_count += 1;
+                GGL_LOGT("Found a colon while parsing Docker URI");
+                continue;
+            }
+            GGL_LOGE("More than three colons found while parsing Docker URI, "
+                     "URI is invalid.");
+            return GGL_ERR_INVALID;
+        }
+        if (uri.data[i - 1] == '@') {
+            if (*at == SIZE_MAX) {
```
AniruddhaKanhere, line 175: I'd much rather maintain a local count instead of the caller setting the correct value of `at` variable to `SIZE_MAX`.

modules/ggl-uri/src/gghttp_uri.c
```diff
@@ -136,3 +138,283 @@ GglError gg_uri_parse(GglArena *arena, GglBuffer uri, GglUriInfo *info) {
...
+                *at = i - 1;
+                GGL_LOGT("Found an @ while parsing Docker URI");
+                continue;
+            }
+            GGL_LOGE("More than one at symbol found while parsing Docker URI, "
```
AniruddhaKanhere, line 180:
```suggestion
            GGL_LOGE("More than one '@' symbol found while parsing Docker URI, "
```

modules/ggl-uri/src/gghttp_uri.c
```diff
@@ -136,3 +138,283 @@ GglError gg_uri_parse(GglArena *arena, GglBuffer uri, GglUriInfo *info) {
...
+            }
+        }
+    }
+    return GGL_ERR_OK;
+}
+
+static GglError parse_docker_registry_segment(
+    GglDockerUriInfo *info,
+    GglBuffer uri,
+    size_t slashes[2],
```
AniruddhaKanhere, line 196: Same here. Use a pointer instead of an array.

AniruddhaKanhere (changes requested):

modules/ggl-uri/src/gghttp_uri.c
archigup: `size_t slashes[static 2]` would require at least 2. Used for compiler options; may warn.

Without the static, its ignored by compiler and just for humans.

modules/ggl-uri/src/gghttp_uri.c
patrzhan: This was written this way since size_t is always positive so there's a warning if I write the for loop the other way as `len >= 0` is always true.

AniruddhaKanhere: Yes, I missed that. Thanks for clarifying.

cookpate:
Do we need to do the requests ourselves to pull an image or can we just parse out that it's a docker uri (i.e. `docker:<docker_uri>`) and pass everything past the first colon to the Docker CLI?

PR #876, opened 2025-05-28
Docker Private ECR Repository Support

cookpate:
*Issue #, if available:*
Support running Docker containers with the workflow established in this document:
https://docs.aws.amazon.com/greengrass/v2/developerguide/run-docker-container.html

*Description of changes:*
* Add unit test to verify Docker URI support.
* Use Docker CLI to pull, check, and remove images
* Support downloading of Docker artifacts during deployments.

*Needs testing:*
* Greengrass IPC from within Docker containers (might just work if passed SVCUID)

*Future work:*
* Stale component Docker artifact removal
* Docker Login for private Docker registries or Private ECR

Marking this as a draft while testing is done. May instead decide to create a branch on the main repository and tag it, later.

### DIFF
--- a/modules/ggdeploymentd/src/deployment_handler.c
+++ b/modules/ggdeploymentd/src/deployment_handler.c
@@ -699,6 +699,7 @@ static GglError get_recipe_artifacts(
         return error;
     }
 
+    bool ecr_logged_in = false;
     for (size_t i = 0; i < artifacts.len; ++i) {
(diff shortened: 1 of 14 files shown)


thomas-roos (commented):
looks good - but should we name it  container instead of docker? May docker is not the thing in the future anymore or we want to be inclusive for podman? just an idea...

archigup (approved):

PR #884, opened 2025-06-18
Fix failure of components with config substitution

archigup:
Handling of config substitution was being handled in the child after fork, with the parent execing bash. This meant that the IPC receive thread was not running.

Swapping the parent and child does not work since systemd treats the first process specially (I was also having some issue with python exiting when recipe runner called waitid). Instead I removed the fork entirely. To minimize this change, a memfd is used to replace stdin of the shell, so exiting script building code works.

### DIFF
--- a/modules/recipe-runner/src/runner.c
+++ b/modules/recipe-runner/src/runner.c
@@ -22,6 +22,7 @@
 #include <ggl/recipe.h>
 #include <ggl/vector.h>
 #include <limits.h>
+#include <sys/mman.h>
 #include <sys/types.h>
 #include <unistd.h>
 #include <stdbool.h>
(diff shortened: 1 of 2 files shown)


modules/recipe-runner/src/runner.c
```diff
@@ -790,34 +801,22 @@ GglError runner(const RecipeRunnerArgs *args) {
...
-
-    pid_t pid = fork();
-
-    if (pid < 0) {
-        GGL_LOGE("Err %d when calling fork.", errno);
+    int script_fd = memfd_create("ggl_component_script", 0);
+    if (script_fd < 0) {
+        GGL_LOGE(
+            "Failed to create memfd for component phase script: %d.", errno
+        );
```
rawalexe, line 808: How does this fix the issue?

archigup: We are no longer forking. We're fully building the script and then `exec`ing it instead of feeding stdin. To reuse existing code that output into the stdin fd, we are swapping the stdin fd for a memfd.

patrzhan (approved):

PR #885, opened 2025-06-20
v2.2.0 pre-release

cookpate:
v2.2.0 pre-release


version
```diff
@@ -1 +1 @@
-2.1.0
+2.2.0-docker
```
rawalexe, line 1: No docker pre-fix needed

rawalexe (approved):

PR #887, opened 2025-06-26
Fix for tes-server failing to vend credentials

rawalexe:
*Issue #, if available:*

*Description of changes:*


archigup:
This will always return true now

archigup (changes requested):
This change removes validation of the svcuid's existence.

archigup (approved):

PR #899, opened 2025-07-17
Update fleet prov to run only once

rawalexe:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/modules/fleet-provisioning/src/entry.c
+++ b/modules/fleet-provisioning/src/entry.c
@@ -31,7 +31,7 @@
 
 #define USER_GROUP (GGL_SYSTEMD_SYSTEM_USER ":" GGL_SYSTEMD_SYSTEM_GROUP)
 
-GglBuffer ggcredentials_path = GGL_STR("/ggcredentials");
+GglBuffer ggcredentials_path = GGL_STR("/var/lib/greengrass/provisioned-cert/");
 
 static GglError start_iotcored(FleetProvArgs *args, pid_t *iotcored_pid) {
     const char *iotcore_d_args[]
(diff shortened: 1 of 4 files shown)


modules/fleet-provisioning/src/entry.c
```diff
@@ -47,6 +47,65 @@ static GglError start_iotcored(FleetProvArgs *args, pid_t *iotcored_pid) {
...
+            GGL_BUF_LIST(
+                GGL_STR("services"),
+                GGL_STR("aws.greengrass.fleet_provisioning"),
+                GGL_STR("configuration"),
+                GGL_STR("claimKeyPath")
+            ),
+            &alloc,
+            &claim_key_path
+        );
+        if (ret != GGL_ERR_OK) {
```
patrzhan, line 70: Only GGL_ERR_NOENTRY implies that the config value isn't set, other errors would mean something else went wrong that isn't the config value not existing.

modules/fleet-provisioning/src/entry.c
```diff
@@ -285,18 +355,39 @@ static GglError update_iot_endpoints(void) {
...
-    GglError ret = fetch_from_db(args);
+    int config_dir;
+    ret = ggl_dir_open(ggcredentials_path, O_RDONLY, false, &config_dir);
+    if (ret != GGL_ERR_OK) {
+        GGL_LOGI(
+            "Could not open %.*s directory.",
+            (int) ggcredentials_path.len,
+            ggcredentials_path.data
+        );
+        return GGL_ERR_FAILURE;
```
patrzhan, line 386: Should return the same error code from ret, instead of changing it to GGL_ERR_FAILURE. Also, should the above log line be error level?

patrzhan (approved):

PR #908, opened 2025-07-24
Run Lychee link checker

felicityzhao9:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,3 +59,13 @@ jobs:
           log-lines = 0
           EOF
           nix build -L .#checks.x86_64-linux.${{ matrix.check }}
+  check-links:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Run Lychee link checker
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -59,3 +59,13 @@ jobs:
...
           EOF
           nix build -L .#checks.x86_64-linux.${{ matrix.check }}
+  check-links:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Run Lychee link checker
+        uses: lycheeverse/lychee-action@v2.4.1
```
archigup, line 69: Should put this in the Nix CI env

.github/workflows/ci.yml
```diff
@@ -59,3 +59,13 @@ jobs:
...
+        with:
+          args: --verbose --require-https --exclude-mail './**/*.md'
```
archigup, line 71: I think we should just check local/relative links. Internet links go down all the time, so we wouldn't make this a required check, and then this would end up perma-red like on other repos.

PR #935, opened 2025-08-27
feat: Add TPM handle support for private key configuration

felicityzhao9:
*Description of changes:*

Enable loading private keys directly from TPM using handle.
Uses OpenSSL OSSL_STORE API with TPM2 provider for hardware-backed keys.


archigup (approved):

thomas-roos:
nice!

PR #948, opened 2025-09-23
Build more compatible deb packages

thomas-roos:
Existing deb packages are only compatible with Ubuntu 24.04. 
This will add alternate and minimal lib package dependency versions to be compatible with
Ubuntu 22.04. 24.04. and Debian 12 + 13.
Also armv7 (arm32) is build.
The built packages are installed in the respective os in a systemd container, checked if 
so libs are missing and if uninstall works.
The static workflow is removed as we do not support this, because of security reasons. 
If you really need a static binary you need to do this yourself.
Exception libzip4 is linked statically as Debian 13 just ships with libzip5, which is not compatible.
Fixes: https://github.com/aws-greengrass/aws-greengrass-lite/issues/819


rawalexe:
I like this PR as it expands the compatibility. But @archigup  Can you also have a look

.github/workflows/ubuntu-24.04-packaging-aarch64.yml
```diff
@@ -90,3 +90,43 @@ jobs:
...
+  install-test:
+    needs: build
+    runs-on: ubuntu-24.04-arm
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: jrei/systemd-ubuntu:22.04
+          - os: jrei/systemd-ubuntu:24.04
+          - os: jrei/systemd-debian:12
```
archigup, line 103: Is this a dependency we want to bring into our supply chain?

archigup:
Could you rebase the branch to commits following format from https://cbea.ms/git-commit/ and update PR title?

.github/workflows/ubuntu-24.04-packaging-aarch64.yml
thomas-roos: We could mitigate using a TAG, but then someone needs to update it.
We are not using this dependency to create the debian packages. 
Changing to

modules/ggl-zip/CMakeLists.txt
thomas-roos, line 1: this is necessary to static link libzip

thomas-roos:
tested (connecting) packages with:
Debian 13 / aka Raspian Trixie arm64
Ubuntu 22.04: x86-64
Ubuntu 24.04: arm64

.github/workflows/deb-packaging-multi-arch.yml
```diff
@@ -0,0 +1,249 @@
...
+          - arch: armv7
+            runner: ubuntu-24.04-arm
+            container: arm32v7/ubuntu:22.04
+            build_arch: arm/v7
+            artifact_suffix: armv7
+            build_type: MinSizeRel
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up QEMU
```
rawalexe, line 54: Maybe like Setup Qemu for arm32

.github/workflows/deb-packaging-multi-arch.yml
```diff
@@ -0,0 +1,249 @@
...
+            platform: linux/amd64
+            os: ubuntu:24.04
+          - arch: x86-64
+            platform: linux/amd64
+            os: debian:12
+          - arch: x86-64
+            platform: linux/amd64
+            os: debian:13
+          - arch: aarch64
+            platform: linux/arm64
```
rawalexe, line 156: Maybe everything below this can be it's own file?

modules/ggl-zip/CMakeLists.txt
```diff
@@ -2,4 +2,4 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-ggl_init_module(ggl-zip LIBS ggl-sdk PkgConfig::libzip)
+ggl_init_module(ggl-zip LIBS ggl-sdk PkgConfig::libzip z)
```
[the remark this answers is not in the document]
thomas-roos: static linking of libzip requires zlib to be linked explicit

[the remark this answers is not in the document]
thomas-roos: added info to the commit message

thomas-roos:
ready for merging or review

rawalexe (approved):
Approved

rawalexe (approved):

PR #956, opened 2025-10-03
Handle stream termination messages in ggipcd

cookpate:
*Issue #, if available:*
* If a terminate stream message is set, ggipcd terminates the entire connection with an error log.

*Description of changes:*
* Gracefully terminate stream when the terminate flag is set.

### DIFF
--- a/modules/ggipcd/src/ipc_subscriptions.c
+++ b/modules/ggipcd/src/ipc_subscriptions.c
@@ -184,3 +184,23 @@ GglError ggl_ipc_release_subscriptions_for_conn(uint32_t resp_handle) {
 
     return GGL_ERR_OK;
 }
+
+void ggl_ipc_terminate_stream(uint32_t resp_handle, int32_t stream_id) {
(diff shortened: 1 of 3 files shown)


modules/ggipcd/src/ipc_subscriptions.c
```diff
@@ -184,3 +185,23 @@ GglError ggl_ipc_release_subscriptions_for_conn(uint32_t resp_handle) {
...
 }
+
+void ggl_ipc_terminate_stream(uint32_t resp_handle, int32_t stream_id) {
+    for (size_t i = 0; i < GGL_COREBUS_CLIENT_MAX_SUBSCRIPTIONS; i++) {
+        uint32_t recv_handle = 0;
+
+        {
+            GGL_MTX_SCOPE_GUARD(&subs_state_mtx);
+
+            if ((subs_resp_handle[i] == resp_handle)
```
archigup, line 196: Don't we need to remove the stream as well?

PR #957, opened 2025-10-03
Update get Configuration as per the Classic behavior 

rawalexe:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/fc_deps.json
+++ b/fc_deps.json
@@ -16,7 +16,7 @@
   },
   "ggl_sdk": {
     "url": "https://github.com/aws-greengrass/aws-greengrass-sdk-lite.git",
-    "rev": "9f9b95363d6de4c68ee5d34ba5c6065b4c1b31fb",
-    "hash": "sha256-tIFIKvZI10WDHBfdFBYAbgHVpNaoUUhzsO8tshk5o3A="
+    "rev": "b83773450bccf6974cded2005792a2400ce5e018",
+    "hash": "sha256-2bvRk23DnGQe9DqbZBDDWt1JzikIbwD4fF8yFr6j+HE="
   }
 }


modules/ggipcd/src/services/config/get_configuration.c
```diff
@@ -89,6 +89,19 @@ GglError ggl_handle_get_configuration(
         return ret;
     }
 
+    // According to the IPC spec, if keyPath has a valid value,
+    //  For MAP values a map without the keyPath leaf is returned.
+    //  For non-MAP values, a map with the keyPath leaf and the value is
+    //  returned.
+    GglObjectType read_type = ggl_obj_type(read_value);
+    if (read_type != GGL_TYPE_MAP && key_path.len > 0) {
+        read_value = ggl_obj_map((GglMap
```
archigup, line 98: this map lives for scope of the if

modules/ggipcd/src/services/config/get_configuration.c
```diff
@@ -89,6 +89,19 @@ GglError ggl_handle_get_configuration(
         return ret;
     }
 
+    // According to the IPC spec, if keyPath has a valid value,
+    //  For MAP values a map without the keyPath leaf is returned.
+    //  For non-MAP values, a map with the keyPath leaf and the value is
+    //  returned.
+    GglObjectType read_type = ggl_obj_type(read_value);
+    if (read_type != GGL_TYPE_MAP && key_path.len > 0) {
```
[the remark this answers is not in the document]
rawalexe: By this point if you have reached the key_path is already valid. So I guess remove the length check

modules/ggipcd/src/services/config/get_configuration.c
rawalexe: Good point

archigup (dismissed):

PR #962, opened 2025-10-17
Add reset-failed on Nucleus and component restart

cookpate:
*Issue #, if available:*
* If components are restarted for any reason, this increments the restart counter towards its burst limit.

*Description of changes:*
* Add global reset-failed on gghealthd startup
* Add reset-failed of a single component on RestartComponent IPC

### DIFF
--- a/modules/gghealthd/src/health.c
+++ b/modules/gghealthd/src/health.c
@@ -7,10 +7,12 @@
 #include "sd_bus.h"
 #include "subscriptions.h"
 #include <assert.h>
+#include <ggl/arena.h>
 #include <ggl/buffer.h>
 #include <ggl/cleanup.h>
 #include <ggl/error.h>
 #include <ggl/exec.h>
+#include <ggl/list.h>
 #include <ggl/log.h>
 #include <ggl/map.h>
 #include <ggl/nucleus/constants.h>
@@ -18,6 +20,7 @@
 #include <ggl/vector.h>
 #include <systemd/sd-bus.h>
 #include <systemd/sd-daemon.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <stdlib.h>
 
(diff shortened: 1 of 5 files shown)


modules/gghealthd/src/health.c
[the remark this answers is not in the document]
rawalexe: You should also probably reset after the restart no?

modules/gghealthd/src/health.c
```diff
@@ -249,8 +236,66 @@ GglError gghealthd_restart_component(GglBuffer component_name) {
     return GGL_ERR_OK;
 }
 
+static bool is_nucleus_component(GglBuffer component_name) {
+    if (ggl_buffer_has_prefix(component_name, GGL_STR("aws.greengrass."))) {
```
aws-kevinrickard: This seems like it would give many false positives, many of our aws-provided generic components like aws.greengrass.StreamManager would match this.

It may be better to use the componentType info from the config to figure out what are the nucleus components vs not

modules/gghealthd/src/health.c
```diff
@@ -249,8 +236,66 @@ GglError gghealthd_restart_component(GglBuffer component_name) {
...
+        }
+        ++reset_count;
+        reset_restart_counters(bus, (char *) qualified_name);
+    }
+    GGL_LOGD("Processed reset-failed for %zu components", reset_count);
+}
+
 GglError gghealthd_init(void) {
     sd_notify(0, "READY=1");
+    reset_failed_components();
```
archigup: shouldn't this be before we are ready?

modules/gghealthd/src/health.c
```diff
@@ -249,8 +236,66 @@ GglError gghealthd_restart_component(GglBuffer component_name) {
     return GGL_ERR_OK;
 }
 
+static bool is_nucleus_component(GglBuffer component_name) {
+    if (ggl_buffer_has_prefix(component_name, GGL_STR("aws.greengrass."))) {
+        return true;
+    }
+    if (ggl_buffer_eq(component_name, GGL_STR("DeploymentService"))) {
```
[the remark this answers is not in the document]
archigup: Does lite report these?

modules/gghealthd/src/sd_bus.c
```diff
@@ -317,3 +319,64 @@ GglError get_lifecycle_state(
     err = get_component_result(bus, unit_path, state);
     return err;
 }
+
+void reset_restart_counters(sd_bus *bus, const char *qualified_name) {
```
aws-kevinrickard: nit: In general I think it's better to return errors when possible, and let the caller of the function decide whether to ignore and continue running or not. e.g. if systemd isn't responding we might want to retry this a couple times before continuing on. Not a big deal in this case though

modules/gghealthd/src/sd_bus.c
```diff
@@ -317,3 +319,64 @@ GglError get_lifecycle_state(
...
+            "Failed to reset failure counter for %s (errno=%d) (name=%s) "
+            "(message=%s)",
+            qualified_name,
+            -ret,
+            error.name,
+            error.message
+        );
+    }
+    GGL_CLEANUP(sd_bus_error_free, error);
+    GGL_CLEANUP(sd_bus_message_unrefp, reply);
```
aws-kevinrickard: Should probably declare cleanups before the if statement, as is done in restart_component

aws-kevinrickard (approved):

PR #963, opened 2025-10-17
Fleet provisioning redesigned

rawalexe:
*Issue #, if available:*

*Description of changes:*


modules/fleet-provisioning/src/config_operations.c
```diff
@@ -0,0 +1,369 @@
...
+GglError ggl_update_iot_endpoints(FleetProvArgs *args) {
+    GglError ret = ggl_gg_config_write(
+        GGL_BUF_LIST(
+            GGL_STR("services"),
+            GGL_STR("aws.greengrass.NucleusLite"),
+            GGL_STR("configuration"),
+            GGL_STR("iotDataEndpoint")
+        ),
+        ggl_obj_buf(ggl_buffer_from_null_term(args->endpoint)),
+        &(int64_t) { 3 }
```
cookpate: Hard-coding a timestamp of 3 makes fleet provisioning potentially non-reentrant. Is it possible for fleet provisioning to fail and restart after this point?

rawalexe: Highly unlikely, it would fail after this point. But can you explain your reasoning here? No other function should be replacing this value other than fleet-prov. And during initial init it's set at 2. So this was to set to always override.

modules/fleet-provisioning/bin/fleet-provisioning.c
```diff
@@ -4,87 +4,82 @@
...
+          OPTION_ARG_OPTIONAL,
+          "AWS IoT Core data endpoint",
           0 },
         { "root-ca-path",
           'r',
           "path",
-          0,
-          "[optional]Path to key for client certificate",
+          OPTION_ARG_OPTIONAL,
+          "Path to key for client certificate",
```
felicityzhao9: I believe root-ca-path means "the path to the Root CA certificate", not for the key path, right?

modules/fleet-provisioning/bin/fleet-provisioning.c
```diff
@@ -94,55 +89,27 @@ static error_t arg_parser(int key, char *arg, struct argp_state *state) {
...
+    GglByteVec iotcored_path = GGL_BYTE_VEC(iotcored_path_buf);
+
+    GglError ret = ggl_binpath_append_name(
+        ggl_buffer_from_null_term(argv[0]), GGL_STR("iotcored"), &iotcored_path
+    );
+    if (ret == GGL_ERR_OK) {
+        args.iotcored_path = (char *) iotcored_path.buf.data;
+    }
+
+    ret = run_fleet_prov(&args);
```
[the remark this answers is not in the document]
rawalexe: We fetch from DB if not provided. It's under config_operations.c

modules/fleet-provisioning/bin/fleet-provisioning.c
rawalexe: Good catch

PR #981, opened 2025-10-30
2025 10 30 main parallel uats

thomas-roos:
Don't forget to add missing secrets AND merge https://github.com/aws-greengrass/aws-greengrass-testing/pull/288

### DIFF
--- a/.github/workflows/uat.yml
+++ b/.github/workflows/uat.yml
@@ -7,13 +7,13 @@ on:
   workflow_dispatch:
     inputs:
       testing_repo:
-        description: 'Testing repository to use as source of tests'
+        description: "Testing repository to use as source of tests"
         required: false
-        default: 'aws-greengrass/aws-greengrass-testing'
+        default: "aws-greengrass/aws-greengrass-testing"
       testing_repo_ref:
-        description: 'Ref (branch) on the testing repo to use'
+        description: "Ref (branch) on the testing repo to use"
         required: false
-        default: 'python_testing'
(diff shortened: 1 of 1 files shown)


.github/workflows/uat.yml
```diff
@@ -36,17 +57,17 @@ jobs:
       - name: Checkout testing repository
         uses: actions/checkout@v4
         with:
-          repository: aws-greengrass/aws-greengrass-testing
-          ref: python_testing
+          repository: ${{ secrets.TESTING_REPO }}
```
felicityzhao9: Duplicated started from line 18

.github/workflows/uat.yml
```diff
@@ -36,17 +57,17 @@ jobs:
...
+          ref: ${{ secrets.TESTING_REPO_REF }}
           path: aws-greengrass-testing
 
       - name: Get temporary AWS credentials
         uses: aws-actions/configure-aws-credentials@v4
         with:
           audience: sts.amazonaws.com
-          aws-region: us-west-2
+          aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
```
felicityzhao9: Since we’re merging to main, let’s make sure we use the same account and region (us-west-2) consistently. There’s no need to store these as secrets.

.github/workflows/uat.yml
thomas-roos: true, might can be reused from first job!?

felicityzhao9 (approved):

PR #1097, opened 2026-04-15
Add PKCS#11 setup instructions

archigup:


docs/PKCS11_SUPPORT.md
```diff
@@ -0,0 +1,119 @@
...
+You need an OpenSSL 3.x provider that bridges PKCS#11. Install the provider and
+your PKCS#11 module. For example, on Ubuntu with SoftHSM:
+
+```bash
+sudo apt install openssl pkcs11-provider softhsm2 opensc
+```
+
+Edit `/etc/ssl/openssl.cnf` to register the pkcs11 provider alongside the
+default provider. In the `[provider_sect]` section, add a `pkcs11` entry:
+
```
felicityzhao9, line 34: Have we already added top-level initialization hook in the OpenSSL config by default? e.g.
```
openssl_conf = openssl_init
[openssl_init]
providers = provider_sect
```
I remember this is the first step to load the provider. Users probably need to add by themselves.

docs/PKCS11_SUPPORT.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+## Greengrass Configuration
+
+Set `privateKeyPath` and `certificateFilePath` to PKCS#11 URIs:
+
+```yaml
+---
+system:
+  privateKeyPath: "pkcs11:token=ggl;object=key;pin-value=1234"
```
felicityzhao9, line 115: Cool so we support PIN in the URI, do we also support in the OpenSSL config? It seems PIN could also been supported through [`pkcs11-module-token-pin`](https://github.com/openssl-projects/pkcs11-provider/blob/main/docs/provider-pkcs11.7.md#pkcs11-module-token-pin). We probably also need to add this instruction if the answer is yes.

docs/PKCS11_SUPPORT.md
```diff
@@ -0,0 +1,119 @@
...
+```bash
+openssl s_client \
+  -connect YOUR_IOT_ENDPOINT:8443 \
+  -key "pkcs11:token=ggl;object=key;pin-value=1234" \
+  -cert "pkcs11:token=ggl;object=cert;pin-value=1234" \
+  -CAfile /path/to/AmazonRootCA1.pem \
+  -brief </dev/null
+```
+
+A successful connection shows `CONNECTION ESTABLISHED` and `Verification: OK`.
```
felicityzhao9, line 137: Not a blocker: shall we also add troubleshooting steps? Ref: https://github.com/archigup/aws-greengrass-lite/blob/53eb2c1094bbf8f2d66ac80289a67b6bc6732bdd/docs/TPM_SUPPORT.md#troubleshooting

felicityzhao9 (approved):

PR #1105, opened 2026-04-24
Suppress ggconfigd subscriber notifications on unchanged writes

### LINKED ISSUE #961: SubscribeToConfigurationUpdate does not work as expected
**Describe the bug**
The subscription may notify [Number of configuration elements +1] times with each deployment that has a config value change. And at least once if there is no change.

**To Reproduce**
- Deploy a component that uses SubscribeToConfigurationUpdate and has more than 1 `DeafultConfiguration` 
- Revise the deployment 
- [Optional] Make some changes to the configuration
- Check how many times the callback is called

**Expected behavior**
The callback for SubscribeToConfigurationUpdate should only be called once per change

**Actual behavior**
The callback for SubscribeToConfigurationUpdate is called more than once per change

### LINKED ISSUE #1017: SubscribeToConfigurationUpdate: a NULL component name is treated as a wildcard instead of the calling component's name
**Describe the bug** 
 
I'm passing NULL for the component name as follows:

```
ret = ggipc_subscribe_to_configuration_update(NULL, GG_BUF_LIST(), config_update_handler, NULL, NULL);
```

Per the documentation, a value of NULL should default to the name of the component making the request: https://docs.aws.amazon.com/greengrass/v2/developerguide/ipc-component-configuration.html#ipc-operation-subscribetoconfigurationupdate. In practice, if you subscribe as NULL, you will receive notifications for configuration changes made to ANY component. So the default in nucleus lite (and/or aws-greengrass-component-sdk) seems to be like a wildcard instead.

**To Reproduce** 

Deploy two components that subscribe with a NULL component name. Modify the configuration of one component. Confirm that both receive a notification.

**Expected behavior**

A component should only receive notifications for changes to its own configuration.

**Actual behavior** 

It gets notified about changes to configuration of any component.

**Environment**

- OS: `Linux ip-172-31-25-129 6.6.111-yocto-standard-00221-g2a17dc587afe`
- Nucleus version: 2.3.0

rawalexe:
- ggconfig_write_value_at_key notified subscribers on every write, even when the new value was byte-identical to the stored one, producing redundant IPC traffic.

- Compare the stored buffer to the incoming one with gg_buffer_eq before notifying. Both are JSON-encoded, so a bytewise compare covers all value types. The update still proceeds to refresh the timestamp; only notify_nested_key is skipped.

## Description

Brief description of the changes in this PR.

## Related Issue

Fixes #1017 #961

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Checklist

- [ ] Code passes all the quality test. Can try `nix flake check -L` locally
      (if applicable)

## Documentation Updates

## Testing

- [x] Existing tests pass
- [x] Added tests to
 - https://github.com/aws-greengrass/aws-greengrass-testing/pull/304

## Additional Notes

Any additional information, context, or screenshots that would be helpful for
reviewers.

### DIFF
--- a/modules/ggconfigd/src/db_interface.c
+++ b/modules/ggconfigd/src/db_interface.c
@@ -842,6 +842,51 @@ GgError ggconfig_write_value_at_key(
         return GG_ERR_OK;
     }
 
+    // Check if the stored value is byte-identical to the incoming one.
+    // Both are JSON-encoded, so buffer compare is sufficient. The update still
+    // proceeds (to refresh the timestamp), but notifications are suppressed
+    // when the value didn't actually change. The pointer returned by
+    // sqlite3_column_text is owned by sqlite and remains valid until the
+    // statement is finalized, which GG_CLEANUP handles at scope exit, so
+    // comparing against it in place avoids an arena allocation and memcpy.
+    bool value_unchanged = false;
+    {
+        sqlite3_stmt *stmt = NULL;
+        int rc = sqlite3_prepare_v2(
+            config_database, GGL_SQL_READ_VALUE, -1, &stmt, NULL
+        );
+        if (rc != SQLITE_OK) {
+            GG_LOGE(
+                "Failed to prepare read-value statement for key id %" PRId64
+                " with rc %d and error %s",
+                last_key_id,
+                rc,
+                sqlite3_errmsg(config_database)
+            );
+        } else {
+            // Register cleanup immediately so the statement is finalized
+            // on any exit path below.
+            GG_CLEANUP(cleanup_sqlite3_finalize, stmt);
+            sqlite3_bind_int64(stmt, 1, last_key_id);
+            rc = sqlite3_step(stmt);
+            if (rc != SQLITE_ROW) {
+                GG_LOGE(
+                    "Failed to read existing value for key id %" PRId64
+                    " with rc %d and error %s",
+                    last_key_id,
+                    rc,
+                    sqlite3_errmsg(config_database)
+                );
+            } else {
+                GgBuffer existing = {
+                    .data = (uint8_t *) sqlite3_column_text(stmt, 0),
+                    .len = (size_t) sqlite3_column_bytes(stmt, 0),
+                };
+                value_unchanged = gg_buffer_eq(existing, *value);
+            }
+        }
+    }
+
     err = value_update(last_key_id, value, timestamp);
     if (err != GG_ERR_OK) {
         GG_LOGE(
@@ -855,6 +900,14 @@ GgError ggconfig_write_value_at_key(
         return err;
     }
     sqlite3_exec(config_database, "END TRANSACTION", NULL, NULL, NULL);
+
+    if (value_unchanged) {
+        GG_LOGD(
+            "key %s value unchanged; skipping subscriber notification",
+            print_key_path(key_path)
+        );
+        return GG_ERR_OK;
+    }
 
     err = notify_nested_key(key_path, ids);
     if (err != GG_ERR_OK) {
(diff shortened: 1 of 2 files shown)


aws-kevinrickard:
Also fixes https://github.com/aws-greengrass/aws-greengrass-lite/issues/961

aws-kevinrickard:
We should update the doc about this problem now that it's being fixed

modules/ggconfigd/src/db_interface.c
```diff
@@ -842,6 +846,24 @@ GgError ggconfig_write_value_at_key(
...
     }
 
+    // Check if the stored value is byte-identical to the incoming one.
+    // Both are JSON-encoded, so buffer compare is sufficient. The update still
+    // proceeds (to refresh the timestamp), but notifications are suppressed
+    // when the value didn't actually change.
+    bool value_unchanged = false;
+    {
+        GgObject existing_value;
+        uint8_t existing_mem[GGCONFIGD_MAX_OBJECT_DECODE_BYTES];
```
aws-kevinrickard: The 9 KB stack buffer isn't needed here. read_value_at_key copies sqlite's
internal buffer into an arena so it can hand back a GgObject the caller owns —
but for a byte-compare you don't need ownership. You just need to look at the
bytes while the sqlite statement is still alive.

Consider inlining the read and comparing sqlite3_column_text against *value directly, with GG_CLEANUP handling sqlite3_finalize at scope exit:

```c
bool value_unchanged = false;
sqlite3_stmt stmt = NULL;
if (sqlite3prepare_v2(config_database, GGL_SQL_READ_VALUE, -1, &stmt, NULL) != SQLITE_OK) {
  // Log error
} else {
   GG_CLEANUP(cleanup_sqlite3_finalize, stmt);
   sqlite3_bind_int64(stmt, 1, last_key_id);
   if (sqlite3_step(stmt) != SQLITE_ROW) {
       // log error 
   } else {
       GgBuffer existing = {
           .data = (uint8_t ) sqlite3column_text(stmt, 0),
           .len  = (size_t) sqlite3_column_bytes(stmt, 0),
       };
       value_unchanged = gg_buffer_eq(existing, value);
   }
}
```
Compared to the current approach this:

* drops 9 KB of stack per call
* removes a memcpy from the hot path
* removes the need for the forward declaration of read_value_at_key
* (bonus) adds error logging for failed sqlite calls, these should be surfaced


The sqlite-owned pointer from sqlite3_column_text is valid until
sqlite3_finalize, which GG_CLEANUP invokes at the end of the block, so the
comparison is safe.

aws-kevinrickard (changes requested):

modules/ggconfigd/src/db_interface.c
```diff
@@ -842,6 +842,51 @@ GgError ggconfig_write_value_at_key(
...
+                "Failed to prepare read-value statement for key id %" PRId64
+                " with rc %d and error %s",
+                last_key_id,
+                rc,
+                sqlite3_errmsg(config_database)
+            );
+        } else {
+            // Register cleanup immediately so the statement is finalized
+            // on any exit path below.
+            GG_CLEANUP(cleanup_sqlite3_finalize, stmt);
```
aws-kevinrickard, line 869: Super-Nit: Cleanup declaration would ideally be placed immediately after prepare to match the pattern (see other examples in file). 

Here there is no cleanup defined for the error path. But in practice I think it shouldn't matter because stmt should be null (no cleanup needed) on the error path anyways

aws-kevinrickard (approved):

PR #1127, opened 2026-06-10
Fix deployment stuck in QUEUED on job re-execution

sahith:
## Problem

Greengrass Lite deployments get permanently stuck in QUEUED state when a CONTINUOUS IoT Job creates a new execution (same job ID, new `executionNumber`) for a device that already completed a previous execution. This happens with dynamic thing groups where devices leave and rejoin the group on MQTT disconnect/reconnect.

The previous dedup logic in `enqueue_job` compared only the job ID:

```c
if (gg_buffer_eq(current_job_id.buf, job_id)) {
    // Always dropped — even if it's a new execution
}
```

Since `current_job_id` is never cleared after terminal status, any re-execution with the same job ID is silently dropped. The job stays QUEUED on the cloud forever, and because IoT Jobs `notify-next` delivers jobs in order, ALL subsequent deployments are also blocked (head-of-line blocking).

## Root Cause

CONTINUOUS jobs targeting dynamic thing groups create new executions (same `jobId`, incremented `executionNumber`, new `queuedAt` timestamp) when a device leaves and rejoins the group. The device-side dedup had no way to distinguish a true duplicate (same notification re-delivered) from a legitimate new execution.

## Fix

Replace the job-ID-only dedup with a timestamp-based comparison that mirrors the Greengrass Classic Nucleus behavior (`LatestQueuedJobs.trackLastKnownJobs`):

- Same job ID AND `queuedAt <= last_queue_at` → duplicate, skip
- Same job ID AND `queuedAt > last_queue_at` → new execution, accept
- Different job ID → always accept

Persist `lastQueueAt` to `ggconfigd` so the dedup survives process restarts. Restore it in `set_jobs_deployment_for_bootstrap` on startup.

## Reproduction

1. Create a CONTINUOUS deployment targeting a thing group
2. Wait for terminal state (SUCCEEDED or FAILED)
3. Kill iotcored to simulate MQTT disconnect
4. Remove and re-add the thing to the group (simulates dynamic thing group re-evaluation)
5. IoT Jobs creates a new execution (`executionNumber: 2`) with the same job ID
6. **Before fix:** Device logs `"Duplicate job document received. Skipping."` — job stays QUEUED forever
7. **After fix:** Device processes the re-execution normally

## Testing

Tested on a podman container running Greengrass Lite built from this branch. All scenarios verified:

| Scenario | Result |
|----------|--------|
| Normal deployment | SUCCEEDED |
| Re-execution after disconnect + group re-add (the bug) | SUCCEEDED (exec#2 processed) |
| Multiple consecutive re-executions (exec#3) | SUCCEEDED |
| Second thing group deployment not head-of-line blocked | SUCCEEDED |
| Process restart — no false dedup on empty queue | Clean startup, no spurious drops |
| SUCCEEDED deployment followed by same-job re-execution | SUCCEEDED (exec#2 processed) |
| True duplicate (same `queuedAt` re-delivered) | Correctly skipped |

## Related

- PR #1123: Fixes job ID overwrite race on concurrent `notify-next` (complementary fix for IN_PROGRESS stuck state)

### DIFF
--- a/modules/ggdeploymentd/src/iot_jobs_listener.c
+++ b/modules/ggdeploymentd/src/iot_jobs_listener.c
@@ -24,12 +24,12 @@
 #include <ggl/core_bus/aws_iot_mqtt.h>
 #include <ggl/core_bus/client.h>
 #include <ggl/core_bus/gg_config.h>
+#include <inttypes.h>
 #include <pthread.h>
 #include <string.h>
 #include <sys/types.h>
 #include <time.h>
 #include <stdbool.h>
-#include <stdint.h>
 #include <stdnoreturn.h>
 
 #define MAX_THING_NAME_LEN 128
@@ -72,6 +72,9 @@ static uint8_t current_job_id_buf[64];
 static GgByteVec current_job_id;
 static uint8_t current_deployment_id_buf[64];
 static GgByteVec current_deployment_id;
+static uint8_t last_queue_job_id_buf[64];
+static GgByteVec last_queue_job_id;
+static int64_t last_queue_at;
 
 static pthread_mutex_t listener_mutex = PTHREAD_MUTEX_INITIALIZER;
 // listener_cond uses CLOCK_MONOTONIC so the flush retry timeout is immune to
@@ -354,14 +357,27 @@ static GgError describe_next_job(void *ctx) {
     return process_job_execution(gg_obj_into_map(*execution));
 }
 
-static GgError enqueue_job(GgMap deployment_doc, GgBuffer job_id) {
+static GgError enqueue_job(
+    GgMap deployment_doc, GgBuffer job_id, int64_t queued_at
+) {
     GgError ret;
     {
         GG_MTX_SCOPE_GUARD(&current_job_id_mutex);
-        if (gg_buffer_eq(current_job_id.buf, job_id)) {
-            GG_LOGI("Duplicate job document received. Skipping.");
+        if (gg_buffer_eq(last_queue_job_id.buf, job_id)
+            && (queued_at <= last_queue_at)) {
+            GG_LOGI(
+                "Duplicate job notification for %.*s "
+                "(queuedAt=%" PRId64 "). Skipping.",
+                (int) job_id.len,
+                job_id.data,
+                queued_at
+            );
             return GG_ERR_OK;
         }
+        last_queue_job_id = GG_BYTE_VEC(last_queue_job_id_buf);
+        GgError append_ret = gg_byte_vec_append(&last_queue_job_id, job_id);
+        assert(append_ret == GG_ERR_OK);
+        last_queue_at = queued_at;
     }
 
     GgByteVec deployment_id_vec = GG_BYTE_VEC((uint8_t[64]) { 0 });
@@ -398,12 +414,17 @@ static GgError process_job_execution(GgMap job_execution) {
     GgObject *job_id = NULL;
     GgObject *status = NULL;
     GgObject *deployment_doc = NULL;
+    GgObject *queued_at_obj = NULL;
     GgError err = gg_map_validate(
         job_execution,
         GG_MAP_SCHEMA(
             { GG_STR("jobId"), GG_OPTIONAL, GG_TYPE_BUF, &job_id },
             { GG_STR("status"), GG_OPTIONAL, GG_TYPE_BUF, &status },
-            { GG_STR("jobDocument"), GG_OPTIONAL, GG_TYPE_MAP, &deployment_doc }
+            { GG_STR("jobDocument"),
+              GG_OPTIONAL,
+              GG_TYPE_MAP,
+              &deployment_doc },
+            { GG_STR("queuedAt"), GG_REQUIRED, GG_TYPE_I64, &queued_at_obj }
         )
     );
     if (err != GG_ERR_OK) {
@@ -441,13 +462,14 @@ static GgError process_job_execution(GgMap job_execution) {
 
     case DSA_ENQUEUE_JOB: {
         if (deployment_doc == NULL) {
-            GG_LOGE(
-                "Job status is queued/in progress, but no deployment doc was given."
-            );
+            GG_LOGE("Job status is queued/in progress, but no deployment doc "
+                    "was given.");
             return GG_ERR_INVALID;
         }
         (void) enqueue_job(
-            gg_obj_into_map(*deployment_doc), gg_obj_into_buf(*job_id)
+            gg_obj_into_map(*deployment_doc),
+            gg_obj_into_buf(*job_id),
+            gg_obj_into_i64(*queued_at_obj)
         );
         break;
     }
@@ -692,6 +714,11 @@ GgError set_jobs_deployment_for_bootstrap(
             return ret;
         }
     }
+
+    last_queue_job_id = GG_BYTE_VEC(last_queue_job_id_buf);
+    GgError ret = gg_byte_vec_append(&last_queue_job_id, job_id);
+    assert(ret == GG_ERR_OK);
+
     return GG_ERR_OK;
 }
 


utsavvyas:
**Blocking — cross-job `last_queue_at` pollution can re-create the stuck-in-QUEUED bug**

`last_queue_at` is a single file-scope global, but in `enqueue_job` the line `last_queue_at = queued_at;` runs for *every* accepted notification — including jobs whose ID differs from `current_job_id`. The skip is gated on `gg_buffer_eq(current_job_id.buf, job_id)`, yet `enqueue_job` never sets `current_job_id` (only `set_current_job` / `set_jobs_deployment_for_bootstrap` do). So an interleaved second job raises the shared watermark, and a later legitimate new execution of the active job is then dropped.

Concrete interleaving:
1. `set_current_job(JobA)` → `current_job_id = A`.
2. JobB notification arrives first → `B != A`, so it's accepted (correct), but `last_queue_at` is set to JobB's `queuedAt` (e.g. 200). `current_job_id` is still A.
3. JobA's new execution (same jobId, new executionNumber) arrives with `queuedAt = 100` → `A == A && 100 <= 200` → **skipped as a duplicate**, even though it's a legitimate re-execution. Deployment stays QUEUED — the exact failure this PR targets, now reachable whenever a second job interleaves.

Note the `&&` short-circuit isn't the issue: for JobB the buffer compare is false so the timestamp compare is skipped and the job is accepted — but the *unconditional* `last_queue_at = queued_at;` after the guard is what pollutes the global.

**Suggested fix:** key the watermark to the job it belongs to (don't reuse `current_job_id`, since the "only update on match" variant still mishandles the job-to-job transition — a newly-active job would inherit the previous job's watermark). Persist the pair so dedup stays per-job across restarts:

```c
static uint8_t last_queue_job_id_buf[64];
static GgByteVec last_queue_job_id;  /* job that last_queue_at belongs to; guarded by current_job_id_mutex */
static int64_t  last_queue_at;       /* guarded by current_job_id_mutex */
```
```c
GG_MTX_SCOPE_GUARD(&current_job_id_mutex);
if (gg_buffer_eq(last_queue_job_id.buf, job_id) && (queued_at <= last_queue_at)) {
    GG_LOGI("Duplicate job notification for %.*s (queuedAt=%" PRId64 "). Skipping.",
            (int) job_id.len, job_id.data, queued_at);
    return GG_ERR_OK;
}
/* ...after a *successful* enqueue, re-key and advance: */
last_queue_job_id = GG_BYTE_VEC(last_queue_job_id_buf);
if (gg_byte_vec_append(&last_queue_job_id, job_id) != GG_ERR_OK) {
    last_queue_job_id.buf.len = 0;  /* id too long: clear so we never falsely match */
}
last_queue_at = queued_at;
```

With a dedicated key, a different job neither pollutes the watermark nor inherits a stale one, and no reset logic in `set_current_job` is required. Please also add a unit test for the cross-job interleave above.

sahith:
> **Blocking — cross-job `last_queue_at` pollution can re-create the stuck-in-QUEUED bug**
> 
> `last_queue_at` is a single file-scope global, but in `enqueue_job` the line `last_queue_at = queued_at;` runs for _every_ accepted notification — including jobs whose ID differs from `current_job_id`. The skip is gated on `gg_buffer_eq(current_job_id.buf, job_id)`, yet `enqueue_job` never sets `current_job_id` (only `set_current_job` / `set_jobs_deployment_for_bootstrap` do). So an interleaved second job raises the shared watermark, and a later legitimate new execution of the active job is then dropped.
> 
> Concrete interleaving:
> 
> 1. `set_current_job(JobA)` → `current_job_id = A`.
> 2. JobB notification arrives first → `B != A`, so it's accepted (correct), but `last_queue_at` is set to JobB's `queuedAt` (e.g. 200). `current_job_id` is still A.
> 3. JobA's new execution (same jobId, new executionNumber) arrives with `queuedAt = 100` → `A == A && 100 <= 200` → **skipped as a duplicate**, even though it's a legitimate re-execution. Deployment stays QUEUED — the exact failure this PR targets, now reachable whenever a second job interleaves.
> 
> Note the `&&` short-circuit isn't the issue: for JobB the buffer compare is false so the timestamp compare is skipped and the job is accepted — but the _unconditional_ `last_queue_at = queued_at;` after the guard is what pollutes the global.
> 
> **Suggested fix:** key the watermark to the job it belongs to (don't reuse `current_job_id`, since the "only update on match" variant still mishandles the job-to-job transition — a newly-active job would inherit the previous job's watermark). Persist the pair so dedup stays per-job across restarts:
> 
> ```c
> static uint8_t last_queue_job_id_buf[64];
> static GgByteVec last_queue_job_id;  /* job that last_queue_at belongs to; guarded by current_job_id_mutex */
> static int64_t  last_queue_at;       /* guarded by current_job_id_mutex */
> ```
> 
> ```c
> GG_MTX_SCOPE_GUARD(&current_job_id_mutex);
> if (gg_buffer_eq(last_queue_job_id.buf, job_id) && (queued_at <= last_queue_at)) {
>     GG_LOGI("Duplicate job notification for %.*s (queuedAt=%" PRId64 "). Skipping.",
>             (int) job_id.len, job_id.data, queued_at);
>     return GG_ERR_OK;
> }
> /* ...after a *successful* enqueue, re-key and advance: */
> last_queue_job_id = GG_BYTE_VEC(last_queue_job_id_buf);
> if (gg_byte_vec_append(&last_queue_job_id, job_id) != GG_ERR_OK) {
>     last_queue_job_id.buf.len = 0;  /* id too long: clear so we never falsely match */
> }
> last_queue_at = queued_at;
> ```
> 
> With a dedicated key, a different job neither pollutes the watermark nor inherits a stale one, and no reset logic in `set_current_job` is required. Please also add a unit test for the cross-job interleave above.

The cross-job pollution scenario requires queuedAt to go backwards in time — a re-execution of JobA arriving with a timestamp lower than JobB's. This can't happen in practice because the IoT Jobs service always sets queuedAt to current wall-clock time at execution creation. A re-execution created after JobB will always have a strictly higher timestamp.

That said, it's good to fix structurally. The updated implementation:

1. Tracks last_queue_job_id separately from current_job_id — eliminates the cross-job concern entirely
2. Keeps dedup state in-memory only (no persistence) — matches GG Classic's LatestQueuedJobs behavior, which also uses Instant.EPOCH as the initial watermark and never persists it

Tested on a live device — all scenarios from the PR description verified including the remove/re-add to thing group re-execution case

utsavvyas (dismissed):

patrzhan:
PR description mentions - "Persist lastQueueAt to ggconfigd so the dedup survives process restarts. Restore it in set_jobs_deployment_for_bootstrap on startup."

I agree with that as part of the fix, but it doesn't seem like this has been added. This still seems relevant to add. The PR restores the job id value in set_jobs_deployment_for_bootstrap but doesn't do anything with regard to last queued at value.

modules/ggdeploymentd/src/iot_jobs_listener.c
```diff
@@ -307,12 +323,17 @@ static GgError process_job_execution(GgMap job_execution) {
...
         job_execution,
         GG_MAP_SCHEMA(
             { GG_STR("jobId"), GG_OPTIONAL, GG_TYPE_BUF, &job_id },
             { GG_STR("status"), GG_OPTIONAL, GG_TYPE_BUF, &status },
-            { GG_STR("jobDocument"), GG_OPTIONAL, GG_TYPE_MAP, &deployment_doc }
+            { GG_STR("jobDocument"),
+              GG_OPTIONAL,
+              GG_TYPE_MAP,
+              &deployment_doc },
+            { GG_STR("queuedAt"), GG_REQUIRED, GG_TYPE_I64, &queued_at_obj }
```
patrzhan, line 427: Should this be GG_REQUIRED? The others are GG_OPTIONAL with checks that happen after this block.

sahith: Yes, intentionally `GG_REQUIRED`. This code path only receives jobs via the $next topic (DescribeJobExecution for $next or notify-next), which only delivers `QUEUED` or `IN_PROGRESS` executions — never terminal states. queuedAt is always present for those. Classic also dereferences it without null-checking (`jobExecutionData.queuedAt.toInstant()`). The terminal states in the status map (SUCCEEDED, FAILED, etc.) are defensive — they shouldn't arrive through this path but are handled gracefully if they do.

sahith:
> PR description mentions - "Persist lastQueueAt to ggconfigd so the dedup survives process restarts. Restore it in set_jobs_deployment_for_bootstrap on startup."
> 
> I agree with that as part of the fix, but it doesn't seem like this has been added. This still seems relevant to add. The PR restores the job id value in set_jobs_deployment_for_bootstrap but doesn't do anything with regard to last queued at value.

Intentional — the commit message was updated to reflect the final design: in-memory only, matching GG Classic's LatestQueuedJobs (which uses `Instant.EPOCH` as the initial watermark and never persists it).

After restart, `last_queue_at = 0` means the dedup is permissive (accepts everything). This is correct because:

1. The $next topic includes the job's current status. Terminal states (SUCCEEDED, FAILED) map to `DSA_DO_NOTHING` — never re-enqueued regardless of dedup.
2. For `QUEUED`/`IN_PROGRESS` jobs, `ggl_deployment_enqueue` → get_matching_deployment provides idempotent protection if the same deployment is already in the queue.
3. set_jobs_deployment_for_bootstrap seeds last_queue_job_id on startup so QoS 1 redeliveries within the same process lifetime are caught after the first accept. Persisting would add a `ggl_gg_config_write` on every accepted notification for zero correctness benefit.

patrzhan (dismissed):

patrzhan (dismissed):

patrzhan (approved):

PR #1129, opened 2026-06-11
Persist and report deployment status across reconnect/reboot

patrzhan:
## Problem
  ggdeploymentd reports deployment status (IN_PROGRESS / SUCCEEDED / FAILED)
  to IoT Jobs with a single fire-and-forget publish. If that publish fails —
  flaky network, or a reboot mid-outage — the update is lost and the cloud-side
  job is left stuck non-terminal. Because nucleus lite tracks work via `$next`,
  that stale job is what $next keeps returning.
  
  ## Change
  Adds a config-backed, single-slot status keeper in ggdeploymentd. The latest
  undelivered status (`{job_id, status}`) is persisted to ggconfigd and re-sent
  when possible:
  
  - **Persist** on publish failure (the single chokepoint in `update_job_to`).
  - **Flush** on MQTT reconnect, on startup (covers reboots — the slot lives in
    ggconfigd, not memory), and on a 60s retry timer while a slot is pending.
  - **Clear** on successful delivery.
  - **Supersede**: a new deployment drops any stale slot (lite only tracks one
    deployment at a time).
  - **Drop on permanent reject**: if the cloud rejects with a non-retryable code
    (e.g. `InvalidStateTransition`, `TerminalStateReached`), clear the slot
    instead of retrying forever. Transport failures and retryable rejects
    (`InternalError`, `RequestThrottled`) keep retrying.
  
  Reuses the existing job-listener thread; no new threads.
  
  ## Commits
  1. Add status_keeper for pending deployment status
  2. Persist deployment status on publish failure
  3. Flush persisted status on reconnect and retry
  4. Clear stale status slot on new deployment
  5. Classify permanent vs retryable job rejects
  
  ## Testing
  - `nix flake check` green on every commit (build, clang-tidy, IWYU, cppcheck,
    tests).
  - Manual end-to-end against real IoT Core: persist → reconnect-flush → clear,
    60s-timer retry, reboot recovery, new-deployment supersede, and
    permanent-reject drop (captured a real `InvalidStateTransition`). Confirmed
    the bug reproduces on `main` (status stuck `IN_PROGRESS`) and is fixed here.

### DIFF
--- a/modules/ggdeploymentd/src/status_keeper.c
+++ b/modules/ggdeploymentd/src/status_keeper.c
@@ -1,0 +1,136 @@
+// aws-greengrass-lite - AWS IoT Greengrass runtime for constrained devices
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "status_keeper.h"
+#include <gg/arena.h>
+#include <gg/buffer.h>
+#include <gg/cleanup.h>
+#include <gg/error.h>
+#include <gg/flags.h>
+#include <gg/log.h>
+#include <gg/map.h>
+#include <gg/object.h>
+#include <ggl/core_bus/gg_config.h>
+#include <pthread.h>
+#include <sys/types.h>
+#include <stdbool.h>
+#include <stddef.h>
+
+// ggconfigd key path for the pending deployment-status slot.
+#define PENDING_STATUS_KEY_PATH \
+    GG_BUF_LIST( \
+        GG_STR("services"), \
+        GG_STR("DeploymentService"), \
+        GG_STR("pendingStatus") \
+    )
+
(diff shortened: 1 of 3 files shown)


modules/ggdeploymentd/src/status_keeper.c
```diff
@@ -0,0 +1,136 @@
...
+#include <gg/object.h>
+#include <ggl/core_bus/gg_config.h>
+#include <pthread.h>
+#include <sys/types.h>
+#include <stdbool.h>
+#include <stddef.h>
+
+// Config slot for the latest undelivered status update: { job_id, status }.
+#define PENDING_STATUS_KEY_PATH \
+    GG_BUF_LIST( \
```
rawalexe: The comment here can be better

patrzhan: Updated

rawalexe (approved):

PR #1130, opened 2026-06-11
recipe2unit: order user-declared dependencies with After=

paabbs:
## Description

`recipe2unit`'s `parse_dependency_type` emits `BindsTo=` (HARD) or `Wants=`
(SOFT) for each user-declared component dependency, but never `After=`. Per
`systemd.unit(5)`, neither directive enforces start ordering, so on a cold boot
systemd starts a dependent component in parallel with the dependency it relies
on.

This matters for `Type=notify` dependencies such as `tes-serverd`
(TokenExchangeService), which binds an ephemeral port, writes it to config, and
only then signals `READY=1`. Without ordering, `recipe-runner` can spawn a
dependent out-of-process component and read a stale port before `tes-serverd`
publishes the current one, baking a dead `AWS_CONTAINER_CREDENTIALS_FULL_URI`
into the component's environment. Any out-of-process component that uses the AWS
SDK container-credentials provider then fails every credential fetch (connection
refused) until it is manually restarted — and because the unit reports
`active (running)`, the failure is silent.

This change adds `After=ggl.<dep>.service` in both the HARD and SOFT branches,
mirroring the existing hardcoded `ggipcd` dependency in the same file. Since
`tes-serverd` is `Type=notify` and signals `READY` only after writing its port,
`After=` alone is sufficient to close the race.

It also changes the TokenExchangeService unit's `[Install] WantedBy=` from
`multi-user.target` to `greengrass-lite.target`, aligning it with every other
greengrass-lite unit (it was the lone outlier, and already declares
`PartOf=greengrass-lite.target`).

## Related Issue

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Checklist

- [x] Code passes all the quality test. `nix flake check -L` passes locally.
- [ ] **Documentation updated** (if applicable) — N/A
- [ ] **Tests added/updated in aws-greengrass-testing** (if applicable) — N/A.
      Verified no unit-file-generation/ordering assertions exist there; this is
      a generation change covered in-tree, and the fix targets a cold-start
      race that isn't deterministically testable in integration.

## Documentation Updates

- [ ] Updated README.md if needed — N/A
- [ ] Updated relevant documentation in `docs/` folder — N/A
- [ ] Requested to update public documentation if applicable — N/A

## Testing

- [x] Existing tests pass (`nix flake check -L`: formatting, clang-tidy, iwyu,
      shellcheck, editorconfig, spelling, cmake-lint, the clang build, and the
      armv6l/musl Pi cross-build all pass).
- [ ] Added tests to aws-greengrass-testing repository — N/A (see above).
- [x] New functionality is covered by a test — extended
      `test_modules/recipe2unit-test` to generate a unit from a recipe with one
      HARD and one SOFT dependency and assert the output contains both the
      dependency directive (`BindsTo=`/`Wants=`) and a matching `After=`.
      Note: per the existing `test_modules` convention these harnesses are
      compiled and linted by `nix flake check` but not auto-executed (not
      CTest-registered); I ran it manually and it exits 0.

## Additional Notes

- **SOFT-dep ordering:** adding `After=` to the `Wants=` branch could form an
  ordering cycle if two components declared mutual dependencies. systemd detects
  ordering cycles at transaction load, drops one `After=` job with a warning,
  and proceeds — so this does not deadlock.
- **Unit-file buffer:** each declared dependency now emits two lines instead of
  one into the fixed `MAX_UNIT_FILE_BUF_SIZE` (2048-byte) buffer in `parser.c`.
  This is bounds-checked and fails safely (returns an error, no overflow), but a
  recipe with a very large number of declared dependencies that previously just
  fit could now exceed it. Happy to bump the buffer if preferred.
- **Bundled `WantedBy` change:** included because, together with `After=`, it is
  part of making TES-dependent components start reliably. I can split it into a
  separate PR if you'd rather keep this one to the `After=` change only.

### DIFF
--- a/test_modules/recipe2unit-test/src/entry.c
+++ b/test_modules/recipe2unit-test/src/entry.c
@@ -2,45 +2,233 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX - License - Identifier : Apache - 2.0
 
+#include <errno.h>
 #include <fcntl.h>
+#include <ftw.h>
 #include <gg/arena.h>
 #include <gg/buffer.h>
+#include <gg/cleanup.h>
 #include <gg/error.h>
 #include <gg/file.h>
 #include <gg/log.h>
 #include <gg/types.h>
 #include <ggl/recipe2unit.h>
+#include <grp.h>
+#include <pwd.h>
 #include <recipe2unit-test.h>
 #include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
 #include <stdbool.h>
 #include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
 
-// For the testing purpose, move the sample recipe.yml to /run/packages/recipes
-// and rename it to recipe-1.0.0.yml
+// Forward declare structure for use in the function below.
+struct stat;
+
+// Placeholder component names kept intentionally generic.
+#define TEST_COMPONENT_NAME "com.example.TestComponent"
+#define TEST_COMPONENT_VERSION "1.0.0"
+#define TEST_HARD_DEP "com.example.HardDep"
+#define TEST_SOFT_DEP "com.example.SoftDep"
+
+// A recipe with one HARD and one SOFT dependency plus a run lifecycle, so the
+// generated run/startup unit file contains the [Unit] dependency directives.
+static const char TEST_RECIPE[]
+    = "---\n"
+      "RecipeFormatVersion: \"2020-01-25\"\n"
+      "ComponentName: " TEST_COMPONENT_NAME "\n"
+      "ComponentVersion: \"" TEST_COMPONENT_VERSION "\"\n"
+      "ComponentDescription: Dependency ordering test component.\n"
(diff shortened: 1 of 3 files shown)


test_modules/recipe2unit-test/src/entry.c
```diff
@@ -2,53 +2,290 @@
...
+#include <errno.h>
 #include <fcntl.h>
+#include <ftw.h>
 #include <gg/arena.h>
 #include <gg/buffer.h>
+#include <gg/cleanup.h>
 #include <gg/error.h>
 #include <gg/file.h>
 #include <gg/log.h>
 #include <gg/types.h>
```
rawalexe, line 14: You can just remove this test file its all good

paabbs:
Superseded by #1131 — moved from my fork to an upstream branch (dev/fix-recipe2unit-after-stale-tes-port) per review feedback, and dropped the recipe2unit-test change per the comment above. Continuing review on #1131.

PR #1141, opened 2026-06-18
Fixes for MQTT loop

archigup:


modules/iotcored/src/tls.c
```diff
@@ -704,15 +717,32 @@ GgError iotcored_tls_write(
...
+                GG_LOGE(
+                    "Write timed out after %d ms; closing connection.",
+                    timeout_ms
+                );
+                ctx->connected = false;
+                return GG_ERR_TIMEOUT;
+            }
+            if (remaining_ms > INT_MAX) {
+                remaining_ms = INT_MAX;
+            }
```
cookpate, line 731: If we are using a monotonic clock and `int` input for timeout, is this condition always false?

archigup: this is remaining time; diff between start time and current

cookpate (approved):

PR #1142, opened 2026-06-22
Various bug fixes and hardening

archigup:


fc_deps.json
```diff
@@ -11,8 +11,8 @@
   },
   "gg_sdk": {
     "url": "https://github.com/aws-greengrass/aws-greengrass-component-sdk.git",
-    "rev": "v1.0.4",
-    "hash": "sha256-es/+Ar6PNbn73kbrn+C5cOM56wJ6N5A3ZRWPwLGDZdw="
+    "rev": "ad5e6534908bf243c168a54eb2ebfc2e4ffd57f6",
```
AniruddhaKanhere, line 14: Not a released version?

modules/iotcored/src/tls.c
```diff
@@ -348,6 +349,12 @@ static GgError do_handshake(char *host, BIO *bio) {
             GG_LOGE("Failed to configure SNI.");
             return GG_ERR_FATAL;
         }
+
+        SSL_set_hostflags(ssl, X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS);
+        if (SSL_set1_host(ssl, host) != 1) {
```
AniruddhaKanhere, line 354: I think we discussed this but an answer here will benefit others.

What happens when the host HTTPS proxy server just has an IP address?

fc_deps.json
archigup: Yup, its main branch. Needs the void cast for var used in assert in sdk so we can have unused var warnings + asserts stripped.

modules/iotcored/src/tls.c
archigup: This depends on the OpenSSL version. >3.3, OpenSSL will detect IP and match against the right field. For 3.0 to 3.2, cert validation will fail. We'd also be sending IP in SNI which should be ignored, but we can clean that up and not send SNI for IP addrs.

modules/ggdeploymentd/src/deployment_queue.h
```diff
@@ -35,4 +35,9 @@ void ggl_deployment_release(GglDeployment *deployment);
 
 GgError deep_copy_deployment(GglDeployment *deployment, GgArena *alloc);
 
+/// Validate a component name.
+///
+/// Names must match `[a-zA-Z0-9._-]+` and be 1-128 characters long
+GgError ggl_validate_component_name(GgBuffer name);
```
AniruddhaKanhere, line 41: Out of curiosity why are all the errors GgError and all the functions are ggl_ (with the extra 'L')?

archigup: GgError is defined in SDK. The types/functions in SDK were renamed ggl -> gg.

AniruddhaKanhere (approved):

PR #1145, opened 2026-06-24
Exit tesd on credential configuration change

sahith:
## Problem

Components using AWS SDKs cache credentials obtained from TES. When credential
configuration changes (e.g., `iotCredEndpoint`, `iotRoleAlias`), the current
in-process URL update in `tesd` does not invalidate SDK-side credential caches
in components. Components continue using stale credentials until they expire (up
to 1 hour).

The `tes_update_cred_url()` function re-reads the endpoint and role alias from
ggconfigd and updates the credential URL in memory — but this only affects
subsequent TES fetches from tesd itself. Component processes with their own SDK
credential caches never see the change.

## Solution

Replace the in-process URL update with `_Exit(0)`. When any credential-relevant
config key changes, tesd exits cleanly. systemd restarts it via
`Restart=always`, and tesd re-reads all configuration fresh on startup.

A follow-up change will add `BindsTo=ggl.core.tesd.service` to tes-serverd,
which will propagate the restart to tes-serverd and all dependent components,
forcing their AWS SDK clients to reinitialize and fetch fresh credentials.

## Changes

- `modules/tesd/src/entry.c`:
  - `cred_config_change_callback()` now logs the changed key and calls `_Exit(0)`
  - All subscriptions consolidated into `subscribe_to_cred_config_changes()`
  - Added subscriptions for 4 system keys (`thingName`, `certificateFilePath`,
    `rootCaPath`, `privateKeyPath`) — unconditional
  - `iotRoleAlias` and `iotCredEndpoint` subscriptions remain conditional on CLI
    override flags (existing behavior)

- `modules/tesd/src/token_service.c`:
  - Removed `tes_update_cred_url()` and `apply_cred_config_changes()` (dead code)
  - Removed `#include <ggl/core_bus/gg_config.h>` (no longer needed)

- `modules/tesd/src/token_service.h`:
  - Removed `tes_update_cred_url()` declaration

## Testing

Verified via endpoint switch deployment on a containerized Greengrass Lite
device (us-east-1 → us-west-2). Relevant journal output:

```
Jun 24 18:28:48 tesd[90]: Credential URL: https://c24redacted.credentials.iot.us-east-1.amazonaws.com/role-aliases/GreengrassCoreTokenExchangeRoleAlias/credentials
Jun 24 18:29:03 tesd[90]: Configuration key 'iotCredEndpoint' changed, restarting tesd.
Jun 24 18:29:03 systemd[1]: ggl.core.tesd.service: Deactivated successfully.
Jun 24 18:29:04 systemd[1]: ggl.core.tesd.service: Scheduled restart job, restart counter is at 1.
Jun 24 18:29:04 systemd[1]: Started ggl.core.tesd.service - core-bus IoT credential provider.
Jun 24 18:29:04 tesd[121]: Credential URL: https://c24redacted.credentials.iot.us-west-2.amazonaws.com/role-aliases/GreengrassCoreTokenExchangeRoleAlias/credentials
```

- tesd detects `iotCredEndpoint` change, logs it, and exits
- systemd restarts tesd ~1 second later (PID 90 → 121)
- On restart, tesd reads fresh config and builds URL with the new endpoint
- Device remains HEALTHY after the switch

### DIFF
--- a/modules/tesd/src/entry.c
+++ b/modules/tesd/src/entry.c
@@ -10,24 +10,128 @@
 #include <gg/types.h>
 #include <ggl/core_bus/gg_config.h>
 #include <ggl/proxy/environment.h>
-#include <string.h>
 #include <tesd.h>
 #include <stdbool.h>
 #include <stdint.h>
+#include <stdlib.h>
 
 static GgError cred_config_change_callback(
     void *ctx, uint32_t handle, GgObject data
 ) {
-    (void) ctx;
     (void) handle;
     (void) data;
+    const char *key = (const char *) ctx;
+    GG_LOGI("Configuration key '%s' changed, restarting tesd.", key);
+    _Exit(0);
+}
 
-    tes_update_cred_url();
+// Subscribe to all config keys that affect credential fetches. When any of
+// these change, tesd exits so systemd restarts it with fresh config.
+//
+// Keys monitored:
+//   iotCredEndpoint     - HTTPS endpoint for the IoT Credential Provider
+//   iotRoleAlias        - IAM role alias used in the credential request URL
+//   thingName           - sent as x-amzn-iot-thingname header in credential req
+//   certificateFilePath - client cert for mTLS to the credential provider
+//   privateKeyPath      - private key for mTLS to the credential provider
+//   rootCaPath          - CA cert for TLS verification of credential provider
+static GgError subscribe_to_cred_config_changes(TesdArgs *args) {
+    GgError ret = ggl_gg_config_subscribe(
+        GG_BUF_LIST(GG_STR("system"), GG_STR("thingName")),
+        cred_config_change_callback,
+        NULL,
+        (void *) "thingName",
+        NULL
+    );
+    if (ret != GG_ERR_OK) {
+        GG_LOGE("Failed to subscribe to thingName changes: %d.", ret);
+        return ret;
+    }
+
+    ret = ggl_gg_config_subscribe(
+        GG_BUF_LIST(GG_STR("system"), GG_STR("certificateFilePath")),
+        cred_config_change_callback,
+        NULL,
+        (void *) "certificateFilePath",
+        NULL
+    );
+    if (ret != GG_ERR_OK) {
+        GG_LOGE("Failed to subscribe to certificateFilePath changes: %d.", ret);
+        return ret;
+    }
+
+    ret = ggl_gg_config_subscribe(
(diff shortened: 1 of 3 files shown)


modules/tesd/src/entry.c
```diff
@@ -10,19 +10,118 @@
...
+// Subscribe to all config keys that affect credential fetches. When any of
+// these change, tesd exits so systemd restarts it with fresh config.
+//
+// Keys monitored:
+//   iotCredEndpoint     - HTTPS endpoint for the IoT Credential Provider
+//   iotRoleAlias        - IAM role alias used in the credential request URL
+//   thingName           - sent as x-amzn-iot-thingname header in credential req
+//   certificateFilePath - client cert for mTLS to the credential provider
+//   privateKeyPath      - private key for mTLS to the credential provider
+//   rootCaPath          - CA cert for TLS verification of credential provider
```
utsavvyas, line 37: How can we differentiate between auth expired and auth change ? We only want it when auth changes right ?

sahith: The subscription fires only when the config value changes in ggconfigd (ggconfigd suppresses notifications when old == new). Credential expiry doesn't change any config value — tesd continues fetching fresh credentials from the same endpoint. Only actual config changes (new endpoint, new role alias, new cert path) trigger the exit.

modules/tesd/src/entry.c
```diff
@@ -10,19 +10,118 @@
...
+        NULL,
+        (void *) "rootCaPath",
+        NULL
+    );
+    if (ret != GG_ERR_OK) {
+        GG_LOGE("Failed to subscribe to rootCaPath changes: %d.", ret);
+        return ret;
+    }
+
+    ret = ggl_gg_config_subscribe(
```
utsavvyas, line 75: nit: Should we actually dont worry about this becuase it should not change right ?

utsavvyas (dismissed):

modules/tesd/src/entry.c
sahith: These values are set once at provisioning and shouldn't change during normal operation — you're right. The subscriptions are defensive for edge cases (cert rotation, re-provisioning) and cost nothing at runtime (no polling, callback only fires on actual value change). Classic's `CredentialRequestHandler` clears its cache on the same 6 keys for the same reason.

archigup (dismissed):

archigup (approved):

PR #1155, opened 2026-07-17
Forward Nucleus config IPC lookups to NucleusLite

patrzhan:
## Description

Components that run on both classic Greengrass and Greengrass Lite need
nucleus configuration (IoT endpoints, proxy settings). On classic this lives
under `aws.greengrass.Nucleus`; on Lite under `aws.greengrass.NucleusLite`,
forcing components to probe both names and fork code paths per runtime.

This PR adds a component-name alias at the IPC boundary: `GetConfiguration`
and `SubscribeToConfigurationUpdate` requests naming `aws.greengrass.Nucleus`
are served from the `aws.greengrass.NucleusLite` configuration tree, while
responses and update events retain the requested name. A component can now use
a single nucleus-agnostic name on both runtimes.

Commits:

1. `Forward Nucleus config IPC lookups to NucleusLite` — the alias for reads
   and subscriptions, with transparent name echo/relabeling via a
   per-subscription context on the IPC subscription wrapper.
2. `Harden IPC subscription binding state` — replaces the parallel
   subscription-state arrays with a slot struct passed directly as the
   core-bus callback context. This also fixes a pre-existing race where an
   event arriving between epoll registration and handle recording failed the
   lookup and closed a fresh subscription.
3. `Centralize Nucleus config IPC alias policy` — a single static alias
   descriptor (`{requested_name, storage_name}`) drives both routing and
   event relabeling; documents the alias and adds a dedicated
   SubscribeToConfigurationUpdate spec.
4. `Clarify Nucleus configuration alias behavior in specs` — req-tags the
   contract (exact, one-way, name-only aliasing) and documents that
   `UpdateConfiguration` and deployment `configurationUpdate` entries use
   component names verbatim.

The alias deliberately does not apply to writes: `UpdateConfiguration` is
caller-scoped (it has no `componentName` parameter), and deployment
configuration updates must target `aws.greengrass.NucleusLite`.

## Related Issue

N/A — requested by an internal component team that runs the same component on
both runtimes.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [x] Code refactoring

## Checklist

- [x] **Tests added/updated in
      [aws-greengrass-testing](https://github.com/aws-greengrass/aws-greengrass-testing/pull/new/dev/nucleusconfigforward)**
      (if applicable)

## Documentation Updates

New/updated specs: `docs/spec/library/ipc_commands/getconfiguration_ipc.md`,
`docs/spec/library/ipc_commands/subscribetoconfigurationupdate_ipc.md` (new),
`docs/spec/executable/ggconfigd.md`, `docs/spec/executable/ggdeploymentd.md`.

## Testing

- [x] Existing tests pass
- [x] Added tests to
      [aws-greengrass-testing repository](https://github.com/aws-greengrass/aws-greengrass-testing/pull/new/dev/nucleusconfigforward)

`test_Component_30_T0` (companion aws-greengrass-testing branch
`dev/nucleusconfigforward`) verifies end to end on a real IoT Core account,
run in a podman systemd container:

- A `GetConfiguration` read via `aws.greengrass.Nucleus` returns the same
  value as a direct `aws.greengrass.NucleusLite` read, and the response
  echoes the requested name.
- Subscribing via `aws.greengrass.Nucleus` is accepted, and a NucleusLite
  configuration change (driven by a cloud deployment merging a probe key) is
  delivered with the event relabeled to `aws.greengrass.Nucleus`.
- No event ever exposes the underlying `aws.greengrass.NucleusLite` name.

Latest run: `1 passed in 72.02s`.

## Additional Notes

- Internal daemons are unaffected: they read config via explicit core-bus
  paths, not the IPC resolver.
- The subscription-state refactor (commit 2) fixes a pre-existing first-event
  race in ggipcd, surfaced while adding the per-subscription callback context;
  it is split into its own commit for reviewability.

### DIFF
--- a/modules/ggipcd/src/ipc_subscriptions.c
+++ b/modules/ggipcd/src/ipc_subscriptions.c
@@ -22,28 +22,50 @@
 
 static_assert(
     GGL_IPC_MAX_SUBSCRIPTIONS <= GGL_COREBUS_CLIENT_MAX_SUBSCRIPTIONS,
-    "IPC max subscriptions exceededs core bus max subscriptions."
+    "IPC subscription maximum exceeds core bus maximum."
 );
 
-static uint32_t subs_resp_handle[GGL_IPC_MAX_SUBSCRIPTIONS];
-static int32_t subs_stream_id[GGL_IPC_MAX_SUBSCRIPTIONS];
-static uint32_t subs_recv_handle[GGL_IPC_MAX_SUBSCRIPTIONS];
-static void *subs_ctx[GGL_IPC_MAX_SUBSCRIPTIONS];
+typedef struct {
+    uint32_t resp_handle;
+    int32_t stream_id;
+    uint32_t recv_handle;
+    GglIpcSubscribeCallback on_response;
+    void *ctx;
+    bool binding;
+    bool close_requested;
+    bool core_closed;
+} IpcSubscriptionSlot;
+
+static IpcSubscriptionSlot subscriptions[GGL_IPC_MAX_SUBSCRIPTIONS];
 static pthread_mutex_t subs_state_mtx = PTHREAD_MUTEX_INITIALIZER;
 
-static GgError init_subs_index(
-    uint32_t resp_handle, int32_t stream_id, void *ctx, size_t *index
+static void reset_sub_slot(IpcSubscriptionSlot *slot) {
+    *slot = (IpcSubscriptionSlot) { 0 };
+}
+
+static GgError reserve_sub_slot(
+    uint32_t resp_handle,
+    int32_t stream_id,
+    GglIpcSubscribeCallback on_response,
+    void *ctx,
+    IpcSubscriptionSlot **result
 ) {
     assert(resp_handle != 0);
+    assert(on_response != NULL);
 
     GG_MTX_SCOPE_GUARD(&subs_state_mtx);
 
     for (size_t i = 0; i < GGL_IPC_MAX_SUBSCRIPTIONS; i++) {
-        if (subs_resp_handle[i] == 0) {
-            subs_resp_handle[i] = resp_handle;
-            subs_stream_id[i] = stream_id;
-            subs_ctx[i] = ctx;
-            *index = i;
+        IpcSubscriptionSlot *slot = &subscriptions[i];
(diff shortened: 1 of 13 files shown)


patrzhan:
Companion integration test PR: aws-greengrass/aws-greengrass-testing#316

modules/ggipcd/src/ipc_subscriptions.c
```diff
@@ -50,83 +74,80 @@ static GgError init_subs_index(
...
+    if (slot->resp_handle == 0) {
+        GG_LOGD("Already released subscription closed.");
+        return;
     }
 
-    GG_LOGD("Already released subscription closed.");
+    if ((slot->recv_handle != 0) && (slot->recv_handle != recv_handle)) {
+        GG_LOGE("Closed an unexpected subscription handle.");
+        return;
+    }
```
rawalexe, line 148: **[Critical]** This mismatched-handle path returns after logging without resetting the slot or setting `close_requested`/`core_closed`. Since no further close will arrive for this slot, it stays occupied permanently — a slot leak in a small fixed-size array. The path should be unreachable given the core-bus fixes ctx per registration, but a permanent leak is the wrong failure mode for a "can't happen" branch in an embedded runtime. Suggest defensively cleaning up after the log, e.g. mark `close_requested`/`core_closed` and `reset_sub_slot(slot)` when `!slot->binding` (mirroring the normal close path below).

patrzhan: Agreed — this branch should be unreachable (ctx is per-registration, and recv_handle is only ever recorded from that same registration), but leaking a slot from a fixed-size array is the wrong failure mode if it ever fires. Fixed as suggested: the mismatch path now marks close_requested/core_closed and releases the slot when not binding, mirroring the normal close path (without adopting the mismatched handle). The sibling mismatch check in subscription_on_response does not need the same treatment: returning an error there makes core-bus close the subscription, which routes back through on_close and cleans up normally — only the close path is terminal. Folded into the subscription-hardening commit.

rawalexe (approved):

PR #1157, opened 2026-07-21
Add SubscribeToIoTCoreConnectionStatus IPC operation

rawalexe:
## Description

Add the `aws.greengrass#SubscribeToIoTCoreConnectionStatus` IPC operation to ggipcd, bridging the existing iotcored `connection_status` core-bus subscription to IPC clients. Components can subscribe and receive the IoT Core MQTT connection status as a stream of `IoTCoreConnectionStatusEvent` messages (`{"connectionStatusEvent":{"status":"CONNECTED"|"DISCONNECTED"}}`), wire-compatible with the Greengrass nucleus implementation and the Java/Python device SDKs.

Design notes:

- The operation response is sent **before** binding the core-bus subscription: iotcored emits the current status immediately on subscription accept, on the core-bus subscription thread, so binding first could emit a stream event before the operation response (an eventstream protocol violation) or lose the first event.
- The current status is delivered as the first stream event immediately after subscribing; subsequent events fire on each connect/disconnect transition.
- No authorization required: informational local-only operation exposing no MQTT topic data, matching the nucleus design (same pattern as `SubscribeToConfigurationUpdate`).
- The stream-event callback tolerates unknown future status values and self-heals (closes the core-bus subscription) if the IPC client connection dies.

## Related Issue

N/A

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Checklist

      (if applicable)

## Documentation Updates

- [x] Updated README.md if needed

## Testing

- [x] Existing tests pass
- [ ] Added tests to

Build verified clean (ggipcd compiles with zero warnings, clang-formatted). The operation spec including the on-the-wire format is documented in `docs/spec/library/ipc_commands/subscribetoiotcoreconnectionstatus_ipc.md`. A matching client API is implemented in aws-greengrass-sdk-lite (separate PR to follow), and the same wire contract was verified end-to-end against the Greengrass nucleus implementation of this operation.

## Additional Notes

The corresponding server-side feature in the classic nucleus is aws-greengrass/aws-greengrass-nucleus#1813. Follow-up candidate (pre-existing, out of scope): `ggl_ipc_bind_subscription` has a recv_handle registration gap that can drop the first event of any core-bus method that emits immediately on accept.

### DIFF
--- a/docs/spec/library/ipc_commands/subscribetoiotcoreconnectionstatus_ipc.md
+++ b/docs/spec/library/ipc_commands/subscribetoiotcoreconnectionstatus_ipc.md
@@ -1,0 +1,124 @@
+# `SubscribeToIoTCoreConnectionStatus` IPC command
+
+The `SubscribeToIoTCoreConnectionStatus` IPC command subscribes a component to
+the status of the AWS IoT Core MQTT connection using the `connection_status`
+method of the `aws_iot_mqtt` core-bus interface provided by `iotcored`.
+
+- [subscribe-to-iot-core-connection-status-1] It will be invoked with the topic
+  `SubscribeToIoTCoreConnectionStatus`.
+- [subscribe-to-iot-core-connection-status-2] It does not require an access
+  control policy in the recipe to work, as it is an informational local-only
+  operation which exposes no MQTT topic data.
+- [subscribe-to-iot-core-connection-status-3] The current connection status is
+  delivered as the first stream event immediately after the subscription is
+  accepted.
+- [subscribe-to-iot-core-connection-status-4] A stream event is delivered on
+  each subsequent connect/disconnect transition of the MQTT connection.
+- [subscribe-to-iot-core-connection-status-5] The operation response is sent
+  before any stream event, per the eventstream protocol requirements.
+
+## Parameters
+
+This operation takes no parameters; the request payload is an empty object.
+
+## Response
+
+- [subscribe-to-iot-core-connection-status-resp-1] On success, returns an empty
+  `SubscribeToIoTCoreConnectionStatusResponse`, followed by a stream of
+  `IoTCoreConnectionStatusEvent` events.
+  - [subscribe-to-iot-core-connection-status-resp-1.1] Each stream event is a
+    map containing a `connectionStatusEvent` key, whose value is a map
+    containing a `status` key.
+  - [subscribe-to-iot-core-connection-status-resp-1.2] `status` is a buffer
+    containing either `CONNECTED` or `DISCONNECTED`.
+- [subscribe-to-iot-core-connection-status-resp-2] On failure, returns a map
+  containing `message`, `_service`, `_message` and `_errorCode`.
+  - [subscribe-to-iot-core-connection-status-resp-2.1] `ServiceError` is
+    returned if the core-bus subscription cannot be established.
+
+## On the wire
+
+```
+structure SubscribeToIoTCoreConnectionStatusRequest {}
+
+structure SubscribeToIoTCoreConnectionStatusResponse {
+    connectionStatusEvent: IoTCoreConnectionStatusEvent
+}
+
+@streaming
+union IoTCoreConnectionStatusEvent {
+    /// The connection status event.
+    connectionStatusEvent: ConnectionStatusEvent
+}
+
+structure ConnectionStatusEvent {
+    /// The connection status.
+    @required
+    status: ConnectionStatus
+}
+
+@enum([
+    {
+        value: "CONNECTED",
+        name: "CONNECTED"
+    },
+    {
+        value: "DISCONNECTED",
+        name: "DISCONNECTED"
+    }
+])
+string ConnectionStatus
+
+operation SubscribeToIoTCoreConnectionStatus {
+    input: SubscribeToIoTCoreConnectionStatusRequest,
+    output: SubscribeToIoTCoreConnectionStatusResponse,
+    errors: [ServiceError]
+}
+```
+
+## Examples
+
+Case 1: Subscribe while the device is connected to AWS IoT Core
+
+- Request: `{}`
+- Response: `{}`
+- First stream event (current status):
+  `{"connectionStatusEvent":{"status":"CONNECTED"}}`
+
+Case 2: MQTT connection is interrupted after subscribing
+
+- Stream event: `{"connectionStatusEvent":{"status":"DISCONNECTED"}}`
+
+Case 3: MQTT connection resumes
+
+- Stream event: `{"connectionStatusEvent":{"status":"CONNECTED"}}`
+
+### RAW on wire:
+
+```shell
+Packet from client:
+  Headers:
+    [:content-type] => String(StrBytes { bytes: b"application/json" })
+    [service-model-type] => String(StrBytes { bytes: b"aws.greengrass#SubscribeToIoTCoreConnectionStatusRequest" })
+    [:message-type] => Int32(0)
+    [:message-flags] => Int32(0)
+    [:stream-id] => Int32(1)
+    [operation] => String(StrBytes { bytes: b"aws.greengrass#SubscribeToIoTCoreConnectionStatus" })
+  Value: {}
+Packet from server (response):
+  Headers:
+    [:content-type] => String(StrBytes { bytes: b"application/json" })
+    [service-model-type] => String(StrBytes { bytes: b"aws.greengrass#SubscribeToIoTCoreConnectionStatusResponse" })
+    [:message-type] => Int32(0)
+    [:message-flags] => Int32(0)
+    [:stream-id] => Int32(1)
+  Value: {}
+Packet from server (stream event):
+  Headers:
+    [:content-type] => String(StrBytes { bytes: b"application/json" })
+    [service-model-type] => String(StrBytes { bytes: b"aws.greengrass#IoTCoreConnectionStatusEvent" })
+    [:message-type] => Int32(0)
+    [:message-flags] => Int32(0)
+    [:stream-id] => Int32(1)
+  Value: {"connectionStatusEvent":{"status":"CONNECTED"}}
+```
--- a/modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
+++ b/modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
@@ -1,0 +1,185 @@
+// aws-greengrass-lite - AWS IoT Greengrass runtime for constrained devices
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "../../ipc_error.h"
+#include "../../ipc_server.h"
+#include "../../ipc_service.h"
+#include "../../ipc_subscriptions.h"
+#include "mqttproxy.h"
+#include <gg/arena.h>
+#include <gg/buffer.h>
+#include <gg/cleanup.h>
+#include <gg/error.h>
+#include <gg/log.h>
+#include <gg/map.h>
+#include <gg/object.h>
+#include <gg/types.h>
+#include <pthread.h>
+#include <sys/types.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+// Inhibit mechanism for the first callback during subscription setup.
+// Serializes correctly because ggipcd dispatches at most one subscribe of this
+// operation at a time; steady-state callbacks bypass via handle+stream compare.
+//
+// All fields below are guarded by setup_mtx.  Sends happen under the mutex so
+// that event ordering on the wire matches the true status ordering.
+static pthread_mutex_t setup_mtx = PTHREAD_MUTEX_INITIALIZER;
+static bool setup_in_progress = false;
+static uint32_t in_progress_handle;
+static int32_t in_progress_stream;
+static int deferred_status = -1; // -1 none, 0 disconnected, 1 connected
+
+/// Wire format: {"connectionStatusEvent": {"status":
+/// "CONNECTED"|"DISCONNECTED"}}
+static GgError send_status_event(
+    uint32_t handle, int32_t stream_id, bool connected
+) {
+    GgBuffer status_str
+        = connected ? GG_STR("CONNECTED") : GG_STR("DISCONNECTED");
+
+    GgMap response = GG_MAP(gg_kv(
+        GG_STR("connectionStatusEvent"),
+        gg_obj_map(GG_MAP(gg_kv(GG_STR("status"), gg_obj_buf(status_str))))
+    ));
+
+    return ggl_ipc_response_send(
+        handle,
+        stream_id,
+        GG_STR("aws.greengrass#IoTCoreConnectionStatusEvent"),
+        response
+    );
+}
+
+static GgError connection_status_callback(
+    void *ctx,
+    GgObject data,
+    uint32_t resp_handle,
+    int32_t stream_id,
+    GgArena *alloc
+) {
+    (void) ctx;
+    (void) alloc;
+
+    if (gg_obj_type(data) != GG_TYPE_BOOLEAN) {
+        GG_LOGE("Unexpected non-boolean connection status object; skipping.");
+        return GG_ERR_OK;
+    }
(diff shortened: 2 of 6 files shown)


modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
```diff
@@ -0,0 +1,120 @@
...
+    // local-only.
+
+    // Send the operation response BEFORE binding the core-bus subscription.
+    // iotcored publishes the current connection status immediately upon
+    // subscription accept, on a separate core-bus subscription thread.
+    // If we bound first, we could:
+    //   (a) violate the eventstream protocol by emitting a stream event
+    //       before the operation response, or
+    //   (b) lose the first event due to the recv_handle registration gap
+    //       inside ggl_ipc_bind_subscription.
```
archigup: hmm we don't know if we will error yet though

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
```diff
@@ -0,0 +1,120 @@
...
+    GgError ret = ggl_ipc_response_send(
+        handle,
+        stream_id,
+        GG_STR("aws.greengrass#SubscribeToIoTCoreConnectionStatusResponse"),
+        (GgMap) { 0 }
+    );
+    if (ret != GG_ERR_OK) {
+        GG_LOGE("Failed to send subscription response.");
+        *ipc_error = (GglIpcError
```
archigup, line 140: setting an IPC error when we failed to do IPC seems off.

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
```diff
@@ -0,0 +1,120 @@
...
+        NULL
+    );
+    if (ret != GG_ERR_OK) {
+        GG_LOGE("Failed to bind connection status subscription after response "
+                "sent; terminating stream.");
+        ggl_ipc_terminate_stream(handle, stream_id);
+        // Return GG_ERR_OK because the success response was already sent.
+        // Returning non-OK here would cause handle_operation (ipc_server.c)
+        // to call send_stream_error, sending a second error frame on an
+        // already-responded stream.
```
archigup: This comment seems useless

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
archigup: We control the callback and we know the stream id, so we can inhibit the first message going out. We'd need to save in a global atomic the current stream id this function is dealing with. When callback runs, if stream id is set, it should update a global atomic instead of responding. when we complete this function, we clear the current state atomic, then check if the callback ran and send the response ourselves if so. the two atomics would need to be seqcst and ordered such that race conditions are handled.

rawalexe: Implemented as described in 1c4dbbaa: bind first; `in_progress_stream` (seq_cst) marks the stream being set up, the callback stores into `deferred_status` instead of responding, and after the response the handler clears `in_progress_stream` then `atomic_exchange`s `deferred_status`. The callback re-checks `in_progress_stream` after its store and does the same exchange, so exactly one side wins and sends the deferred event regardless of interleaving.

rawalexe: Reordered back to bind-before-respond — bind failure now takes the normal `*ipc_error` path with no response committed. 1c4dbbaa

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
rawalexe: Agreed — when `ggl_ipc_response_send` itself fails we now just log and return the error without setting `*ipc_error`. 1c4dbbaa

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
rawalexe: Removed along with the terminate-stream path it described. 1c4dbbaa

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
```diff
@@ -0,0 +1,161 @@
...
+    (void) alloc;
+
+    if (gg_obj_type(data) != GG_TYPE_BOOLEAN) {
+        GG_LOGE("Unexpected non-boolean connection status object; skipping.");
+        return GG_ERR_OK;
+    }
+
+    bool connected = gg_obj_into_bool(data);
+
+    if (atomic_load(&in_progress_stream) == stream_id) {
```
archigup: Stream IDs are per-connection counters (every client's first subscribe is stream 1), so `stream_id` alone doesn't identify a stream — everywhere else in ggipcd streams are keyed by `(resp_handle, stream_id)` (e.g. `subs_resp_handle`/`subs_stream_id` in `ipc_subscriptions.c`, `ggl_ipc_terminate_stream`).

Concretely: while component B's subscribe (stream 1) is in setup, a steady-state callback for component A's already-established subscription (also stream 1, different handle) matches this check, gets deferred into the shared `deferred_status`, and A's transition event is swallowed (B's handler sends it on B's stream instead). A permanently misses a CONNECTED/DISCONNECTED transition. The error paths that clear `deferred_status` can likewise wipe another stream's deferred event.

The guard needs to include the response handle — e.g. pack handle+stream into a single `atomic_uint_fast64_t`, or compare both.

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
```diff
@@ -0,0 +1,161 @@
...
+        atomic_store(&in_progress_stream, -1);
+        atomic_store(&deferred_status, -1);
+        GG_LOGW(
+            "Failed to send subscription response (error %s).", gg_strerror(ret)
+        );
+        return ret;
+    }
+
+    atomic_store(&in_progress_stream, -1);
+    int s = atomic_exchange(&deferred_status, -1);
```
archigup: There's an out-of-order window between clearing `in_progress_stream` (line 148) and draining `deferred_status` here:

1. callback defers status X during setup and returns
2. handler stores `in_progress_stream = -1`
3. a new callback fires with status Y, sees the guard cleared, sends Y directly
4. handler then exchanges `deferred_status`, gets X, and sends it *after* Y

The client ends with stale status X as its last event — for a status stream the last event is the one that matters. An atomics-only design can't close this window, because ordering correctness requires the send itself to be inside the mutual exclusion. Given `ggl_ipc_response_send` already serializes on a global mutex, a small `pthread_mutex` protecting {guard check + send} in the callback and {response send + guard clear + deferred drain} in the handler would be simpler and actually correct — the lock-free dance isn't buying anything here.

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
```diff
@@ -0,0 +1,161 @@
...
+#include <gg/types.h>
+#include <stdatomic.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+// Inhibit mechanism for the first callback during subscription setup.
+// Serializes correctly because ggipcd dispatches at most one subscribe of this
+// operation at a time; steady-state callbacks bypass via stream_id compare.
+static atomic_int_fast32_t in_progress_stream = ATOMIC_VAR_INIT(-1);
```
archigup: Nit: `ATOMIC_VAR_INIT` is deprecated in C17 and removed in C23, and isn't used elsewhere in this repo. Plain initialization is valid for static storage duration:

```c
static atomic_int_fast32_t in_progress_stream = -1;
static atomic_int deferred_status = -1;
```

archigup (dismissed):
Verified the concurrency assumptions against the codebase: ggipcd does dispatch all IPC operations on a single epoll thread, and core-bus client subscription callbacks are serialized on a single subscription thread, so the two-thread store/re-check/exchange handoff is internally sound. However there are two real issues with the guard design, noted inline.

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
rawalexe: Good catch — the guard now compares both `resp_handle` and `stream_id`, so another connection's stream 1 no longer matches. 166b27cb

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
rawalexe: Agreed — switched to a `pthread_mutex` (`setup_mtx`) guarding {guard check + send} in the callback and {response send + guard clear + deferred drain} in the handler, so sends happen inside the mutual exclusion and wire order matches status order. Atomics dropped entirely (which also moots the ATOMIC_VAR_INIT nit). Lock order is consistent: setup_mtx is always acquired before ggipcd's internal send locks, and ggl_ipc_response_send never calls back into this file. 166b27cb

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
rawalexe: Gone along with the atomics — plain statics under the mutex now. 166b27cb

cookpate (dismissed):

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
```diff
@@ -0,0 +1,177 @@
...
+
+// Inhibit mechanism for the first callback during subscription setup.
+// Serializes correctly because ggipcd dispatches at most one subscribe of this
+// operation at a time; steady-state callbacks bypass via handle+stream compare.
+//
+// All fields below are guarded by setup_mtx.  Sends happen under the mutex so
+// that event ordering on the wire matches the true status ordering.
+static pthread_mutex_t setup_mtx = PTHREAD_MUTEX_INITIALIZER;
+static uint32_t in_progress_handle;
+static int32_t in_progress_stream = -1; // -1 = no setup in progress
```
patrzhan: The `-1` sentinel can collide with a client-supplied stream id. Dispatch only rejects `stream_id == 0` (`ipc_server.c`, `handle_operation`), so a client can send `:stream-id` = -1. Setup then completes by resetting `in_progress_stream` back to -1, which now equals the live stream id, and `in_progress_handle` is never cleared. Every subsequent callback for that subscription matches the setup-in-progress test and is swallowed into `deferred_status`, so the client never receives an event. Rejecting non-positive stream ids at dispatch, or tracking setup-in-progress with a separate flag instead of a sentinel value, would remove the aliasing.

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
```diff
@@ -0,0 +1,177 @@
...
+        s = deferred_status;
+        deferred_status = -1;
+        in_progress_stream = -1;
+
+        if (s >= 0) {
+            send_ret = send_status_event(handle, stream_id, s == 1);
+        }
+    }
+
+    if (s >= 0 && send_ret != GG_ERR_OK) {
```
patrzhan, line 177: If the deferred first event fails to send, this logs and returns `GG_ERR_OK`, so the subscription stays alive but the client never learns the current status until the next transition. That contradicts spec item 3 (current status delivered as the first stream event). It is also a different policy from the steady-state callback above, which returns the error so core-bus closes the subscription. Suggest picking one intentional failure policy for this stream; terminating it here seems right, since a failed write to the client usually means the connection is gone anyway.

patrzhan (commented):
Did a deep pass over this, including the surrounding ggipcd/core-bus/iotcored code paths. The deferral design and the response-before-bind ordering analysis are well reasoned. Three issues I think should be fixed in this PR are inline (stream-id sentinel collision, deferred-event failure handling, spec doc inconsistencies). Additional observations below; their root causes are mostly outside this handler, so they need a scoping decision rather than necessarily a fix here.

**A. The recv_handle registration gap you flagged bites this operation on every call.** `ggl_subscribe` adds the fd to the subscription epoll (`core-bus/src/client_subscribe.c`) before `ggl_ipc_bind_subscription` stores the recv handle (`ipc_subscriptions.c`), and iotcored emits the current status synchronously on accept (`iotcored/src/bus_server.c`). If the subscription thread wins, `subscription_on_response` fails the lookup and core-bus closes the subscription, so the client gets a successful response and a silently dead stream. For existing operations the window is never exercised because the first event only arrives with real traffic; here it is exercised on every subscribe, on targets where preemption in that window is plausible. Suggest treating the `ipc_subscriptions` fix as a prerequisite for this PR rather than a follow-up, or explicitly accepting the race in a comment.

**B. Lock-order inversion (pre-existing pattern, noting for the record).** IPC connection cleanup holds the socket-pool mutex through `on_release` (`socket_handle.c`) and then waits on the core-bus client pool mutex via `release_client_subscriptions` -> `ggl_client_sub_close`. The callback thread holds the core-bus pool mutex (`client_subscribe.c`, callback runs under `ggl_socket_handle_protected`) while sending, which takes the IPC pool mutex inside `ggl_ipc_response_send`. That is an ABBA pair: cleanup takes IPC pool then core-bus pool, callback takes core-bus pool then IPC pool. The sibling subscription handlers have the same shape, so I do not think it blocks this PR, but it probably deserves its own issue against the locking design.

**E. Every new subscriber causes duplicate events to all existing subscribers.** `rpc_get_status` delivers the initial status via `iotcored_mqtt_status_update_send` (`iotcored/src/bus_server.c`), which broadcasts to every registered status handle (`subscription_dispatch.c`), not just the newly accepted one. So each new subscription sends a spurious duplicate current-status event to every existing subscriber with no transition having occurred, which violates spec item 4 as written. Either iotcored should respond only to the new handle, or the doc should state at-least-once/duplicate-possible semantics.

**G. Transitions during setup are coalesced.** `deferred_status` keeps only the latest value, so multiple transitions inside the setup window collapse to one event, which is at odds with item 4's per-transition wording. The window is tiny and latest-status is arguably the right semantic for a status stream; if you keep the behavior, suggest adding a line to the spec that events may be coalesced during subscription setup.

docs/spec/library/ipc_commands/subscribetoiotcoreconnectionstatus_ipc.md
```diff
@@ -0,0 +1,124 @@
...
+  - [subscribe-to-iot-core-connection-status-resp-2.1] `ServiceError` is
+    returned if the core-bus subscription cannot be established.
+
+## On the wire
+
+```
+structure SubscribeToIoTCoreConnectionStatusRequest {}
+
+structure SubscribeToIoTCoreConnectionStatusResponse {
+    connectionStatusEvent: IoTCoreConnectionStatusEvent
```
patrzhan, line 45: This sketch gives the response a `connectionStatusEvent` member, but resp-1, the examples, and the RAW capture all show the response as an empty `{}` with the event as a separate stream message, and the generated model in aws-iot-device-sdk-java-v2 (`SubscribeToIoTCoreConnectionStatusResponse.java`) is empty too. The response structure here should be `{}` with the streaming union expressed separately. Related doc nit: line 35 says failures contain `message`, `_service`, `_message` and `_errorCode`, but `send_stream_error` only serializes `_message` and `_errorCode` (the wording looks inherited from `getconfiguration_ipc.md`, which has the same inaccuracy).

modules/ggipcd/src/services/mqttproxy/subscribe_to_iot_core_connection_status.c
rawalexe: This is expected

patrzhan (approved):

PR #1164, opened 2026-08-06
Enabled greengrass components to receive direct message by opting for RECEIVE_ONLY mode

tushar-aws:
## Description

Adds an on-device path for AWS IoT Core direct messages. A component can
subscribe over IPC with `subscriptionMode=RECEIVE_ONLY`; iotcored registers the
topic filter for local routing only and never issues a cloud SUBSCRIBE. Any
message that reaches the device on a matching filter — including direct messages
addressed to the device by client id, which arrive over the existing connection
with no subscription — is delivered to the subscribing component. A message
matching no subscriber is now dropped and logged instead of silently discarded.

Changes by component:

- **ggipcd (`SubscribeToIoTCore`)**: new optional `subscriptionMode`.
  `RECEIVE_ONLY` maps to a routing-only (virtual) subscription; absent, null,
  and `SUBSCRIBE_AND_RECEIVE` keep the existing cloud-subscribe path unchanged;
  any other value is rejected with `InvalidArgumentsError` (fail closed rather
  than silently creating a paid cloud subscription). `qos` is ignored for
  `RECEIVE_ONLY` since there is no cloud subscription to configure.

- **iotcored + core bus (`aws_iot_mqtt` `subscribe`)**: new `virtual` flag.
  A virtual registration sends no cloud SUBSCRIBE, sends no UNSUBSCRIBE on close,
  is not re-subscribed on MQTT reconnect, and does not keep a cloud subscription
  alive for a filter it shares with a cloud subscription. Inbound dispatch
  matches virtual and cloud slots identically (by topic filter), which is what
  lets a virtual slot receive direct messages. A message matching no slot is
  dropped with a warning log.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Checklist

- [ ] Code passes all the quality test. Can try `nix flake check -L` locally
- [ ] **Documentation updated** (if applicable)
      (if applicable)

## Documentation Updates

- [ ] Updated relevant documentation in `docs/` folder

## Testing

- [x] Existing tests pass
- [ ] Added tests to

## Additional Notes

Any additional information, context, or screenshots that would be helpful for
reviewers.

### DIFF
--- a/modules/iotcored/src/bus_server.c
+++ b/modules/iotcored/src/bus_server.c
@@ -141,9 +141,13 @@ static GgError rpc_subscribe(void *ctx, GgMap params, uint32_t handle) {
         return GG_ERR_INVALID;
     }
 
-    bool virtual = false;
+    bool is_virtual = false;
     if (gg_map_get(params, GG_STR("virtual"), &val)) {
-        virtual = gg_obj_into_bool(*val);
+        if (gg_obj_type(*val) != GG_TYPE_BOOLEAN) {
+            GG_LOGE("Subscribe received invalid arguments.");
+            return GG_ERR_INVALID;
+        }
+        is_virtual = gg_obj_into_bool(*val);
     }
 
     for (size_t i = 0; i < topic_filter_count; i++) {
@@ -168,13 +172,13 @@ static GgError rpc_subscribe(void *ctx, GgMap params, uint32_t handle) {
     }
 
     GgError ret = iotcored_register_subscriptions(
-        topic_filters, topic_filter_count, handle, qos
+        topic_filters, topic_filter_count, handle, qos, is_virtual
     );
     if (ret != GG_ERR_OK) {
         return ret;
     }
 
-    if (!virtual) {
+    if (!is_virtual) {
         ret = iotcored_mqtt_subscribe(topic_filters, topic_filter_count, qos);
         if (ret != GG_ERR_OK) {
             iotcored_unregister_subscriptions(handle, false);
--- a/modules/iotcored/src/subscription_dispatch.c
+++ b/modules/iotcored/src/subscription_dispatch.c
@@ -34,6 +34,10 @@ static uint8_t sub_topic_filters[IOTCORED_MAX_SUBSCRIPTIONS]
                                 [AWS_IOT_MAX_TOPIC_SIZE];
 static uint32_t handles[IOTCORED_MAX_SUBSCRIPTIONS];
 static uint8_t topic_qos[IOTCORED_MAX_SUBSCRIPTIONS];
+/// True if the slot is a virtual registration: routed to like any other slot,
+/// but with no cloud subscription behind it (never subscribed, unsubscribed,
+/// or re-subscribed with AWS IoT Core).
+static bool sub_virtual[IOTCORED_MAX_SUBSCRIPTIONS];
 static pthread_mutex_t mtx = PTHREAD_MUTEX_INITIALIZER;
 
 static uint32_t mqtt_status_handles[IOTCORED_MAX_SUBSCRIPTIONS];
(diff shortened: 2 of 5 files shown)


modules/iotcored/src/bus_server.c
```diff
@@ -143,6 +143,10 @@ static GgError rpc_subscribe(void *ctx, GgMap params, uint32_t handle) {
 
     bool virtual = false;
```
muboheng: be careful using virtual here in cpp virtual itself is a keyword, but since this is C, it should be fine?

modules/iotcored/src/subscription_dispatch.h
```diff
@@ -11,8 +11,16 @@
...
+/// `virtual` marks the registration as having no cloud subscription: matching
+/// inbound messages are still routed to `handle`, but the filter is never
+/// subscribed, unsubscribed, or re-subscribed with AWS IoT Core.
 GgError iotcored_register_subscriptions(
-    GgBuffer *topic_filters, size_t count, uint32_t handle, uint8_t qos
+    GgBuffer *topic_filters,
+    size_t count,
+    uint32_t handle,
+    uint8_t qos,
+    bool virtual
```
muboheng: virtual in header file is more dangerous. nit: switch to is_virtual

modules/iotcored/src/subscription_dispatch.c
```diff
@@ -151,6 +170,14 @@ void iotcored_mqtt_receive(const IotcoredMsg *msg) {
             );
         }
     }
+
+    if (!matched) {
+        GG_LOGW(
```
muboheng, line 175: @archigup would this be an expensive operation for gg lite? do we need to demote this or its fine to keep as is. For a scenario where the core device receives lots of unsolicited messages

AniruddhaKanhere: I'd say this is fine. A device should not receive anything it did not subscribe to. If it is, then something is wrong - either we should be subscribed, or the cloud incorrectly thinks that we are.

modules/iotcored/src/subscription_dispatch.c
```diff
@@ -87,61 +96,71 @@ GgError iotcored_register_subscriptions(
...
     return GG_ERR_NOMEM;
 }
 
-// NOLINTNEXTLINE(readability-function-cognitive-complexity)
+/// True if another slot holds a cloud subscription to the same topic filter.
+/// Virtual slots don't count, as they have no cloud subscription.
+static bool cloud_sub_shares_filter(size_t index) {
+    for (size_t i = 0; i < IOTCORED_MAX_SUBSCRIPTIONS; i++) {
+        if ((i == index) || (topic_filter_len[i] == 0) || sub_virtual[i]) {
+            continue;
```
AniruddhaKanhere, line 111: Should we check for handle[index] == handle[i] here too?

Same handle subscribing to same topic multiple times would be left out and will not be unsubscribed from the cloud in the above check. Or am i overthinking this scenario? :)

AniruddhaKanhere (dismissed):
This looks good to me. Thank you @tushar-aws!

modules/iotcored/src/subscription_dispatch.c
tushar-aws: No, this scenario will not exists because two duplicate slots are never both alive at the same moment when the check runs. The loop looks at them one at a time, and erases each before moving on.

modules/iotcored/src/subscription_dispatch.h
tushar-aws: Done

modules/iotcored/src/bus_server.c
tushar-aws: switched it to `virtual` <> `is_virtual`.

AniruddhaKanhere (approved):

patrzhan (approved):

allanli4 (approved):

PR #1167, opened 2026-08-12
Add regex support to recipe platform matching

rawalexe:
## Description

Component recipes may specify a platform attribute value as a slash-delimited regular expression such as `/windows|linux/`, per the Greengrass v2 recipe reference. Matching is whole-string, following the Java nucleus, where `PlatformHelper.isAttributeSatisfied` calls `String.matches`.

This adds a `ggl-regex` module and routes the runtime, os and architecture checks in `manifest_selection()` through it. Exact values, `*` and `all` behave exactly as before.

The engine is a hand-written Thompson NFA rather than libc `regex.h`, because this codebase performs no dynamic allocation. One `regcomp` plus `regexec` plus `regfree` measures 101 heap allocations against a zero-allocation baseline, and the tree has no direct allocator calls and no allocating libc functions. Working memory comes from a caller-provided `GgArena` backed by a stack buffer.

The `/.../` delimiting and the `*` wildcard stay in `ggl-recipe`, since they are recipe conventions rather than regex ones. `ggl-regex` takes a bare pattern.

## Related Issue

N/A

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Checklist

- [x] **Tests added/updated in
      (if applicable)
       - https://github.com/aws-greengrass/aws-greengrass-testing/pull/325

## Documentation Updates

## Testing

- [x] Existing tests pass
- [x] Added tests to

## Additional Notes

Verified on the committed state: full `-DBUILD_TESTING=ON` build clean with no errors or warnings, `ctest` 12/12 passing (the new `ggl-regex-inline-test` target takes it up from 11), 42 regex tests and 20 recipe tests, `clang-format` clean, `clang-tidy` with no unsuppressed warnings, and `cspell` clean.

Supported syntax: literals, `.`, the quantifiers `*` `+` `?`, alternation, grouping (no capture semantics), bracket expressions with ranges and leading-caret negation, escapes of non-alphanumeric characters, and redundant `^` and `$` anchors at the start and end of a pattern.

Java syntax this engine does not implement now **fails closed** with a parse error rather than silently changing meaning: `{n,m}` counted quantifiers, the shorthand classes `\d` `\w` `\s`, any escape of an alphanumeric character, a backslash inside a bracket expression, lookahead, non-greedy quantifiers and named groups. This closes a real gap raised in review: `a{2}` previously matched the literal string `a{2}` and `\d` matched the letter `d`. Both are valid Java, so a recipe using them was not malformed to its author, and it could select a manifest they did not intend.

Two divergences remain and are documented rather than rejected, since both are permissive rather than dangerous: mid-pattern `^` and `$` are literals, and the POSIX rule making a leading `]` in a bracket expression a literal (`[]a]`) is accepted here but rejected by Java. All of the above is recorded in docs/RECIPE_SUPPORT_CHANGES.md.

The caps for support are 256 bytes of pattern, 512 instructions and 16 nesting levels, and every unsupported or malformed pattern fails closed, so a bad pattern can never select a manifest.

Worth knowing for review: the Greengrass cloud compiles recipe platform patterns with Java's `Pattern` during dependency resolution, so a genuinely malformed pattern is rejected before the device is ever consulted. A UAT carrying `/[/` failed its deployment with `Unclosed character class near index 0` from the cloud, not from this engine. The device-side fail-closed path therefore matters mainly for patterns the cloud accepts but this engine cannot honor, which is exactly the set the rejections above now cover.

Deliberately out of scope, present in the Java nucleus but not implemented here: the `any` alias, `all` on architecture, and the guard rejecting templates that do not begin with an alphanumeric. Each would change existing matching
behavior.

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

### DIFF
--- a/docs/RECIPE_SUPPORT_CHANGES.md
+++ b/docs/RECIPE_SUPPORT_CHANGES.md
@@ -68,8 +68,9 @@ supported as mentioned in the
 
   **NOT supported (vs Java regex / PCRE):**
 
-  - `\d`, `\w`, `\s` and their negations (use `[0-9]`, `[a-zA-Z0-9_]`, `[ \t\n]`
-    instead)
+  - `\d`, `\w`, `\s` and their negations are **rejected** (return parse error).
+    Use `[0-9]`, `[a-zA-Z0-9_]`, or a literal space instead. (Platform attribute
+    values never contain tabs or newlines, so a single space suffices for `\s`.)
   - `{` and `{n,m}` counted quantifiers are **rejected** (not matched literally)
   - Lookahead and lookbehind assertions (`(?=...)`, `(?<=...)`)
   - Non-greedy quantifiers (`*?`, `+?`)
--- a/docs/RECIPE_SUPPORT_CHANGES.md
+++ b/docs/RECIPE_SUPPORT_CHANGES.md
@@ -72,6 +72,8 @@ supported as mentioned in the
     Use `[0-9]`, `[a-zA-Z0-9_]`, or a literal space instead. (Platform attribute
     values never contain tabs or newlines, so a single space suffices for `\s`.)
   - `{` and `{n,m}` counted quantifiers are **rejected** (not matched literally)
+  - Backslash escapes inside bracket expressions are **rejected** — there is no
+    escape processing inside `[...]`; use POSIX literal rules instead
   - Lookahead and lookbehind assertions (`(?=...)`, `(?<=...)`)
   - Non-greedy quantifiers (`*?`, `+?`)
   - Named capture groups (`(?<name>...)`)
--- a/docs/RECIPE_SUPPORT_CHANGES.md
+++ b/docs/RECIPE_SUPPORT_CHANGES.md
@@ -70,6 +70,7 @@ supported as mentioned in the
 
   - `\d`, `\w`, `\s` and their negations (use `[0-9]`, `[a-zA-Z0-9_]`, `[ \t\n]`
     instead)
+  - `{` and `{n,m}` counted quantifiers are **rejected** (not matched literally)
   - Lookahead and lookbehind assertions (`(?=...)`, `(?<=...)`)
   - Non-greedy quantifiers (`*?`, `+?`)
   - Named capture groups (`(?<name>...)`)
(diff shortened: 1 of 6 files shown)


patrzhan:
In `nfa_parse_atom` (modules/ggl-regex/src/regex.c), `{` falls through to the literal-byte branch, so `a{2}` matches the literal string "a{2}" instead of "aa". The escape branch has the same property: it emits whatever byte follows the backslash, so `\d` compiles to a literal `d`, and `/\d+/` matches "ddd" here but digits on the Java nucleus.

Both are valid Java regex syntax, so a recipe using them silently changes meaning instead of failing closed. That works against the stated invariant that a bad pattern must never select a manifest: to the recipe author these patterns are not malformed, they are Java syntax this engine does not implement, and they can select or skip a manifest in a way the author did not intend.

Two suggestions:

1. Reject unescaped `{` with `GG_ERR_INVALID`, same as a bare quantifier.
2. In the escape branch, only accept escapes of non-alphanumeric characters, so `\d`, `\w`, `\s` fail closed instead of matching the letter.

At minimum, RECIPE_SUPPORT_CHANGES.md should call out the literal-match behavior. It currently lists `\d` and friends as unsupported but does not say they silently match the letter instead of erroring.

patrzhan:
Escapes are not handled inside bracket expressions. The bracket scan in `nfa_parse_atom` and the matcher in `nfa_class_match` both treat `\` as an ordinary byte, so:

- `[\]]` ends the class at the first `]`, leaving a class that matches only a backslash plus a stray literal `]` at the top level. The Java nucleus reads `[\]]` as "match a literal `]`".
- `[a\-z]` parses `\-z` as a range from `\` (0x5C) to `z` (0x7A). The Java nucleus reads it as the three literals `a`, `-`, `z`.

Same category as the `{n,m}` comment: silent divergence from Java rather than fail-closed. Supporting `\` inside brackets or rejecting it are both reasonable; either way a line in RECIPE_SUPPORT_CHANGES.md would help (e.g. "no escape processing inside bracket expressions").

patrzhan (approved):

